PR #2, opened 2015-05-23
Added _metis.pyx and _api.pxd

OrkoHunter:
This adds _metis.pyx and _api.pxd files.

### DIFF
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -1,0 +1,441 @@
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
+from contextlib import contextmanager
+from os import dup, dup2, fdopen
+from sys import stdout
+from tempfile import TemporaryFile
+from _api cimport *
+from ._types import MetisError
+
+__all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
+
+
+@contextmanager
+def redirect(source, target):
+    """Temporarily redirect operations on file ``source`` to file ``target``.
+    """
+    source.flush()
+    fd = source.fileno()
+    with fdopen(dup(fd), source.mode) as source2:
+        dup2(target.fileno(), fd)
+        try:
+            yield
+        finally:
+            source.flush()
+            dup2(source2.fileno(), fd)
+
+
+cdef void* checked_malloc(idx_t size) except NULL:
+    cdef void* ptr = PyMem_Malloc(size)
+    if ptr == NULL:
+        raise MemoryError()
+    return ptr
+
+
+cdef idx_t* convert_idx_array(array) except NULL:
+    """Convert a list of ints to a C array of idx_t's.
+    """
+    cdef idx_t size = len(array)
+    cdef idx_t *_array = <idx_t*> checked_malloc(sizeof(idx_t) * size)
+    cdef idx_t i
+    try:
+        for i from 0 <= i < size:
+            _array[i] = array[i]
+    except:
+        PyMem_Free(_array)
+        raise
+    return _array
+
+
+cdef real_t* convert_real_array(array) except NULL:
+    """Convert a list of floats to a C array of real_t's.
+    """
+    cdef idx_t size = len(array)
+    cdef real_t *_array = <real_t*> checked_malloc(sizeof(real_t) * size)
+    cdef idx_t i
+    try:
+        for i from 0 <= i < size:
+            _array[i] = array[i]
+    except:
+        PyMem_Free(_array)
+        raise
+    return _array
+
+
+cdef convert_options(options, idx_t *_options):
+    """Convert a MetisOptions object to a C array.
+    """
+    METIS_SetDefaultOptions(_options)
+    if options is None:
+        return
+
+    _options[<idx_t> METIS_OPTION_PTYPE]     = options.ptype
+    _options[<idx_t> METIS_OPTION_OBJTYPE]   = options.objtype
+    _options[<idx_t> METIS_OPTION_CTYPE]     = options.ctype
+    _options[<idx_t> METIS_OPTION_IPTYPE]    = options.iptype
+    _options[<idx_t> METIS_OPTION_RTYPE]     = options.rtype
+    _options[<idx_t> METIS_OPTION_NCUTS]     = options.ncuts
+    _options[<idx_t> METIS_OPTION_NSEPS]     = options.nseps
+    _options[<idx_t> METIS_OPTION_NUMBERING] = options.numbering
+    _options[<idx_t> METIS_OPTION_NITER]     = options.niter
+    _options[<idx_t> METIS_OPTION_SEED]      = options.seed
+    _options[<idx_t> METIS_OPTION_MINCONN]   = options.minconn
+    _options[<idx_t> METIS_OPTION_NO2HOP]    = options.no2hop
+    _options[<idx_t> METIS_OPTION_CONTIG]    = options.contig
+    _options[<idx_t> METIS_OPTION_COMPRESS]  = options.compress
+    _options[<idx_t> METIS_OPTION_CCORDER]   = options.ccorder
+    _options[<idx_t> METIS_OPTION_PFACTOR]   = options.pfactor
+    _options[<idx_t> METIS_OPTION_UFACTOR]   = options.ufactor
+    _options[<idx_t> METIS_OPTION_DBGLVL]    = options.dbglvl
+
+
+cdef void convert_graph(xadj, adjncy, idx_t *nvtxs_ptr, idx_t **_xadj_ptr,
+                        idx_t **_adjncy_ptr) except *:
+    """Convert a list-based graph structure to C arrays.
+    """
+    cdef idx_t nvtxs
+    cdef idx_t *_xadj = NULL
+    cdef idx_t *_adjncy = NULL
+    cdef idx_t i, j, k, l
+    try:
+        _xadj = convert_idx_array(xadj)
+        nvtxs = len(xadj) - 1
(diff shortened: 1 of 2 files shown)


_metis.pyx
```diff
@@ -0,0 +1,437 @@
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
+from pxd._api cimport *
```
SanketDG: Shouldn't this be
```python
from _api.pxd cimport *
```
or
```python
from _api cimport *
```

ysitu: I understand that I wrote this. But since there is only one .pxd file, there is no need to create a directory for it.

_metis.pyx
```diff
@@ -0,0 +1,437 @@
...
+from contextlib import contextmanager
+from os import dup, dup2, fdopen
+from sys import stdout
+from tempfile import TemporaryFile
+
+__all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
+
+
+@contextmanager
+def redirect(file, tmp):
```
ysitu: Please add a docstring.

_metis.pyx
OrkoHunter: I believe there was some `relative import from non-package` if the pxd file was in the top level.
Also, considering other large add-ons, I think it'd be a good practice.

PS. I'm trying to remove the directory in a short while.

OrkoHunter: No errors! I'll remove the pxd directory.

_metis.pyx
```diff
@@ -0,0 +1,439 @@
...
+    """Provides a duplicate of file namespace as tmp
```
OrkoHunter: @ysitu This might have a better docstring line. Please suggest if so.

OrkoHunter:
@ysitu I think I am facing an ambiguous error. Earlier I said that renaming `gklib.c` into `gklib.cc` worked fine. But as of now, If I fetch and combine all the 4 branches from [orkohunter/networkx-metis](https://github.com/orkohunter/networkx-metis) i.e. `master`, `setup`, `wrapper` and `pyx-and-pxd` and do 
`$ python setup.py build_clib` 
followed by 
`$ python setup.py build_ext --inplace`, 
then this error is generated.
```sh
test/ python                                                        
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import metis
  File "<stdin>", line 1, in <module>
  File "metis.py", line 19, in <module>
    from _metis import node_nd, part_graph, compute_vertex_separator
ImportError: ./_metis.so: undefined symbol: libmetis__rpqSeeTopVal
```
Can you please look at it sometime?

ysitu:
See inline comments in #4.

OrkoHunter:
@ysitu Thanks for the fix in `setup.py`. That helped building the .cc files. But, I've encountered another issue while building the library.

Related to `gklib.cc`, it seems that the declared datatypes are not being recognized. Here's a small gist
```sh
src/libmetis/gklib.cc:16:27: error: ‘ikv_t’ does not name a type
     bool operator()(const ikv_t &lhs, const ikv_t &rhs) const {
                           ^
src/libmetis/gklib.cc: In member function ‘bool ikvi_t_less::operator()(const int&, const int&) const’:
src/libmetis/gklib.cc:17:35: error: request for member ‘key’ in ‘lhs’, which is of non-class type ‘const int’
         return std::make_pair(lhs.key, lhs.val) < std::make_pair(rhs.key, rhs.val);
                                   ^
src/libmetis/gklib.cc: At global scope:
src/libmetis/gklib.cc:22:27: error: ‘uvw_t’ does not name a type
     bool operator()(const uvw_t &lhs, const uvw_t &rhs) const {
                           ^
```

and here's one example of the error in `sort.cc`

```sh
In file included from /usr/include/c++/4.8/bits/stl_algobase.h:67:0,
                 from /usr/include/c++/4.8/algorithm:61,
                 from src/GKlib/sort.cc:5:
/usr/include/c++/4.8/bits/stl_iterator.h:347:5: note: template<class _IteratorL, class _IteratorR> bool std::operator<(const std::reverse_iterator<_Iterator>&, const std::reverse_iterator<_IteratorR>&)
     operator<(const reverse_iterator<_IteratorL>& __x,
     ^
/usr/include/c++/4.8/bits/stl_iterator.h:347:5: note:   template argument deduction/substitution failed:
In file included from /usr/include/c++/4.8/algorithm:62:0,
                 from src/GKlib/sort.cc:5:
/usr/include/c++/4.8/bits/stl_algo.h:89:18: note:   ‘gk_fkv_t’ is not derived from ‘const std::reverse_iterator<_Iterator>’
    else if (*__a < *__c)
                  ^
```

Are we still missing something in `setup.py`?

ysitu:
You probably need to move your C includes above the anonymous namespace.

_metis.pyx
ysitu: What this function does is to temporarily redirect operations on a file to another one. The use case is to capture the output to `stdout` by METIS into temporary files so that whether an error occurs, an exception can be raised with an error message.

_metis.pyx
```diff
@@ -0,0 +1,439 @@
...
+    file.flush()
+    fd = file.fileno()
+    with fdopen(dup(fd), file.mode) as file2:
+        dup2(tmp.fileno(), fd)
+        try:
+            yield
+        finally:
+            file.flush()
+            dup2(file2.fileno(), fd)
+            tmp.seek(0)
```
ysitu: Now I sorta find it a bad design to put `seek(0)` here. It has nothing to do with redirection and should be moved out.

_metis.pyx
```diff
@@ -0,0 +1,439 @@
...
+        _xadj[nvtxs] = k
+
+        nvtxs_ptr[0] = nvtxs
+        _xadj_ptr[0] = _xadj
+        _adjncy_ptr[0] = _adjncy
+
+
+cdef void check_result(int result, msg) except *:
+    if result != METIS_OK:
+        from .types import MetisError
```
ysitu: Move this to the top.

OrkoHunter:
The `std::greater<>()` also seems to have problem with these data types.

```sh
In file included from /usr/include/c++/4.8/algorithm:62:0,
                 from src/GKlib/sort.cc:5:
/usr/include/c++/4.8/bits/stl_algo.h: In instantiation of ‘void std::__insertion_sort(_RandomAccessIterator, _RandomAccessIterator) [with _RandomAccessIterator = gk_ckv_t*]’:
/usr/include/c++/4.8/bits/stl_algo.h:2211:62:   required from ‘void std::__final_insertion_sort(_RandomAccessIterator, _RandomAccessIterator) [with _RandomAccessIterator = gk_ckv_t*]’
/usr/include/c++/4.8/bits/stl_algo.h:5453:47:   required from ‘void std::sort(_RAIter, _RAIter) [with _RAIter = gk_ckv_t*]’
src/GKlib/sort.cc:176:27:   required from here
/usr/include/c++/4.8/bits/stl_algo.h:2137:13: error: no match for ‘operator<’ (operand types are ‘gk_ckv_t’ and ‘gk_ckv_t’)
    if (*__i < *__first)
             ^
/usr/include/c++/4.8/bits/stl_algo.h:2137:13: note: candidates are:
In file included from /usr/include/c++/4.8/utility:70:0,
                 from /usr/include/c++/4.8/algorithm:60,
                 from src/GKlib/sort.cc:5:
/usr/include/c++/4.8/bits/stl_pair.h:220:5: note: template<class _T1, class _T2> bool std::operator<(const std::pair<_T1, _T2>&, const std::pair<_T1, _T2>&)
     operator<(const pair<_T1, _T2>& __x, const pair<_T1, _T2>& __y)
     ^
/usr/include/c++/4.8/bits/stl_pair.h:220:5: note:   template argument deduction/substitution failed:
In file included from /usr/include/c++/4.8/algorithm:62:0,
                 from src/GKlib/sort.cc:5:
/usr/include/c++/4.8/bits/stl_algo.h:2137:13: note:   ‘gk_ckv_t’ is not derived from ‘const std::pair<_T1, _T2>’
    if (*__i < *__first)
             ^
```

ysitu:
You need to define comparison functors similar to the ones you have in `gklib.cc`.

OrkoHunter:
Yes I tried it. This leads to
```
src/GKlib/sort.cc: In member function ‘bool {anonymous}::gk_ikv_t_greater::operator()(const gk_ikv_t&, const gk_ikv_t&) const’:
src/GKlib/sort.cc:24:18: error: no match for ‘operator>’ (operand types are ‘const gk_ikv_t’ and ‘const gk_ikv_t’)
         return a > b;
                  ^
```

ysitu:
You cannot just use `>`. That is never defined. You need to look at the original `sort.c` and replicate the comparison logic.

_metis.pyx
OrkoHunter: @ysitu I've just removed it for now. Where should I move `target.seek(0)` out to? It has been used at 3 places.

ysitu: Move it to places immediately before where the target files are read.

_metis.pyx
```diff
@@ -0,0 +1,439 @@
...
+from _api cimport *
+from ._types import MetisError
+
+__all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
+
+
+@contextmanager
+def redirect(source, target):
+    """Provides a temporary duplicate for ``source`` file to raise
+    exceptions on error.
```
ysitu: Temporarily redirect operations on file `source` to file `target`.

PR #3, opened 2015-05-23
Added python wrappers

OrkoHunter:
This adds python wrappers and tests.

### DIFF
--- a/networkx/addons/metis/enums.py
+++ b/networkx/addons/metis/enums.py
@@ -1,0 +1,118 @@
+import enum
+
+__all__ = ['MetisPType', 'MetisObjType', 'MetisCType', 'MetisIPType',
+           'MetisRType', 'MetisNumbering', 'MetisDbgLvl', 'MetisRStatus']
+
+
+@enum.unique
+class MetisPType(enum.IntEnum):
+    """Partitioning method.
+
+        ==== =================================
+        rb   Multilevel recursive bisectioning
+        kway Multilevel `k`-way partitioning
+        ==== =================================
+    """
+
+    default, rb, kway = range(-1, 2)
+
+
+@enum.unique
+class MetisObjType(enum.IntEnum):
+    """Type of objective.
+
+        === =======================================
+        cut Edge-cut minimization
+        vol Total communication volume minimization
+        === =======================================
+    """
+
+    default, cut, vol = range(-1, 2)
+
+
+@enum.unique
+class MetisCType(enum.IntEnum):
+    """Catching scheme to be used during coarsening.
+
+        ==== ==========================
+        rm   Random matching
+        shem Sorted heavy-edge matching
+        ==== ==========================
+    """
+
+    default, rm, shem = range(-1, 2)
+
+
+@enum.unique
+class MetisIPType(enum.IntEnum):
+    """Algorithm used during initial partitioning.
+
+        ====== ======================================================
+        grow   Grow a bisection using a greedy strategy
+        random Compute a bisection at random followed by a refinement
+        edge   Derive a separator from an edge cut
+        node   Grow a bisection using a greedy node-based strategy
+        ====== ======================================================
+    """
+
+    default, grow, random, edge, node = range(-1, 4)
+
+
+@enum.unique
+class MetisRType(enum.IntEnum):
+    """Algorithm used for refinement.
+
+        ========= ======================================
+        fm        FM-based cut refinement
+        greedy    Greedy-based cut and volume refinement
+        sep2sided Two-sided node FM refinement
+        sep1sided One-sided node FM refinement
+        ========= ======================================
+    """
+
+    default, fm, greedy, sep2sided, sep1sided = range(-1, 4)
+
+
+@enum.unique
+class MetisNumbering(enum.IntEnum):
+    """Numbering scheme is used for the adjacency structure of a graph or the
+    element-node structure of a mesh.
+
+        ==== =================================
+        zero C-style zero-based numbering
+        one  Fortran-style one-based numbering
+        ==== =================================
+    """
+
+    default, zero, one = range(-1, 2)
+
+
+@enum.unique
+class MetisDbgLvl(enum.IntEnum):
+    """Amount of progress/debugging information will be printed during the
+    execution of the algorithms. Can be combined by bit-wise OR.
+
+        ========== ======================================================
+        info       Print various diagnostic messages
+        time       Perform timing analysis
+        coarsen    Display various statistics during coarsening
+        refine     Display various statistics during refinement
+        ipart      Display various statistics during initial partitioning
+        moveinfo   Display detailed information about vertex moves during
+                   refinement
+        sepinfo    Display information about vertex separators
+        conninfo   Display information related to the minimization of
+                   subdomain connectivity
+        contiginfo Display information related to the elimination of
+                   connected components
+        ========== ======================================================
+    """
+
+    default = -1
+    (info, time, coarsen, refine, ipart, moveinfo, sepinfo, conninfo,
+     contiginfo, memory) = [1 << x for x in range(9) + [11]]
+
+
+@enum.unique
+class MetisRStatus(enum.IntEnum):
+    ok, error_input, error_memory, error = 1, -2, -3, -4
--- a/networkx/addons/metis/metis.py
+++ b/networkx/addons/metis/metis.py
@@ -1,0 +1,309 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2015 ysitu <ysitu@users.noreply.github.com>
+# All rights reserved
+
+"""
+Wrappers of METIS graph partitioning functions.
+"""
+
+import contextlib
+import decorator
+import itertools
+import sys
+
+import networkx as nx
+
+from networkx.addons.metis import _metis
+from networkx.addons.metis import types
+__all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
+           'MetisOptions']
+
+MetisOptions = types.MetisOptions
+
+
+@contextlib.contextmanager
+def _zero_numbering(options):
+    """Temporarily force zero-based numbering."""
+    if options:
+        numbering = options.numbering
+        options.numbering = MetisNumbering.zero
+    try:
+        yield
+    finally:
+        if options:
+            options.numbering = numbering
+
+
+def _convert_graph(G):
+    """Convert a graph to the numbered adjacency list structure expected by
+    METIS.
+    """
+    index = dict(zip(G, list(range(len(G)))))
+    xadj = [0]
+    adjncy = []
+    for u in G:
+        adjncy.extend(index[v] for v in G[u])
+        xadj.append(len(adjncy))
+    return xadj, adjncy
+
+
+def _convert_exceptions(convert_type, catch_types=None):
+    """Decorator to convert types of exceptions
+
+    Parameters
+    ----------
+    convert_type : subclass of Exception
+        Target type to convert to.
+
+    catch_types : tuple of subclasses of Exception, optional
+        Source types whose instances are to be caught and converted. If None,
+        all instances of Exception will be caught and converted.
+
+    Returns
+    -------
+    _convert_exceptions : function
+        Function that performs exception type conversion.
+
+    Example
+    -------
+    Decorate functions like this::
+
+        @_convert_exceptions(nx.NetworkXError, (ValueError,))
+        def function():
+            pass
+    """
+    @decorator.decorator
+    def _convert_exceptions(func, *args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except catch_types or () as e:
+            exc = e
+        except Exception as e:
+            if catch_types is not None:
+                raise
+            exc = sys.exec_info()
+        six.reraise(convert_type, convert_type(exc[1]), exc[2])
+    return _convert_exceptions
+
+
+@nx.utils.not_implemented_for('directed')
+@nx.utils.not_implemented_for('multigraph')
+@_convert_exceptions(nx.NetworkXError, (ValueError, TypeError, types.MetisError))
+def node_nested_dissection(G, weight='weight', options=None):
+    """Compute a node ordering of a graph that reduces fill when the Laplacian
+    matrix of the graph is LU factorized. The algorithm aims to minimize the
+    sum of weights of vertices in separators computed in the process.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+        A graph.
+
+    weight : object, optional
+        The data key used to determine the weight of each node. If None, each
+        node has unit weight. Default value: 'weight'.
+
+    options : MetisOptions, optional
+        METIS options. If None, the default options are used. Default value:
+        None.
+
+    Returns
+    -------
+    perm : list of nodes
+        The node ordering.
+
+    Raises
+    ------
+    NetworkXError
+        If the parameters cannot be converted to valid METIS input format, or
+        METIS returns an error status.
+    """
+    if len(G) == 0:
+        return []
+
+    vwgt = [G.node[u].get(weight, 1) for u in G]
+    if all(w == 1 for w in vwgt):
+        vwgt = None
+
+    xadj, adjncy = _convert_graph(G)
+
+    with _zero_numbering(options):
+        perm = _metis.node_nd(xadj, adjncy, vwgt, options)[0]
+
+    nodes = list(G)
+    perm = [nodes[i] for i in perm]
+
+    return perm
+
+
+@nx.utils.not_implemented_for('directed')
+@nx.utils.not_implemented_for('multigraph')
+@_convert_exceptions(nx.NetworkXError, (ValueError, TypeError, types.MetisError))
+def partition(G, nparts, node_weight='weight', node_size='size',
+              edge_weight='weight', tpwgts=None, ubvec=None, options=None,
+              recursive=False):
+    """Partition a graph using multilevel recursive bisection or multilevel
+    multiway partitioning.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+        An undirected graph.
+
+    nparts : int
+        Number of parts to partition the graph. It should be at least 2.
+
+    node_weight : object, optional
+        The data key used to determine the weight of each node. If None, each
+        node has unit weight. Default value: 'weight'.
+
+    node_size : object, optional
+        The data key used to determine the size of each node when computing the
+        total communication volumne. If None, each node has unit size. Default
+        value: 'size'
+
+    edge_weight : object, optional
+        The data key used to determine the weight of each edge. If None, each
+        edge has unit weight. Default value: 'weight'.
+
+    tpwgts : list of lists of floats, optional
+        The target weights of the partitions and the constraints. The target
+        weight of the `i`th partition and the `j`th constraint is given by
+        :samp:`tpwgts[i][j]` (the numbering for both partitions and
+        constraints starts from zero). For each constraint of the
+        :samp:`tpwgts[][]` entries must be 1.0 (i.e.,
+        `\sum_i \text{tpwgts}[i][j] = 1.0`)
+
+        If None, the graph is equally divided among the partitions. Default
+        value: None.
+
+    ubvec : list of floats, optional
+        The allowed load imbalance tolerance for each constraint. For the`i`th
+        and the `j`th constraint, the allowed weight is the
+        :samp:`ubvect[j] * tpwgts[i][j]` fraction of the `j`th constraint's
+        total weight. The load imbalances must be greater 1.0.
+
+        If None, the load imbalance tolerance is 1.001 if there is exactly one
+        constraint or 1.01 if there are more. Default value: None.
+
+    options : MetisOptions, optional.
+        METIS options. If None, the default options are used. Default value:
+        None.
+
+    recursive : bool, optional
+        If True, multilevel recursive bisection is used. Otherwise, multileve
+        multilevel multiway partitioning is used. Default value: False.
+
+    Returns
+    -------
+    objval : int
+        The edge-cut or the total communication volume of the partitioning
+        solution. The value returned depends on the partitioining's objective
+        function.
+
+    parts : lists of nodes
+        The partitioning.
+
+    Raises
+    ------
+    NetworkXNotImplemented
+        If the graph is directed or is a multigraph.
+
+    NetworkXError
+        If the parameters cannot be converted to valid METIS input format, or
+        METIS returns an error status.
+    """
+    if nparts < 1:
+        raise nx.NetworkXError('nparts is less than one.')
+    if nparts == 1:
+        return 0, [list(G)]
+
+    if len(G) == 0:
+        return 0, [[] for i in range(nparts)]
+
+    xadj, adjncy = _convert_graph(G)
+
+    vwgt = [G.node[u].get(node_weight, 1) for u in G]
+    if all(w == 1 for w in vwgt):
+        vwgt = None
+
+    vsize = [G.node[u].get(node_size, 1) for u in G]
+    if all(w == 1 for w in vsize):
+        vsize = None
+
+    adjwgt = [G[u][v].get(edge_weight, 1) for u in G for v in G[u]]
+    if all(w == 1 for w in adjwgt):
+        adjwgt = None
+
+    if tpwgts is not None:
+        if len(tpwgts) != nparts:
+            raise nx.NetworkXError('length of tpwgts is not equal to nparts.')
+        ncon = len(tpwgts[0])
+        if any(len(tpwgts[j]) != ncon for j in range(1, nparts)):
+            raise nx.NetworkXError(
+                'lists in tpwgts are not of the same length.')
+        if ubvec is not None and len(ubvec) != ncon:
+            raise nx.NetworkXError(
+                'ubvec is not of the same length as tpwgts.')
+        tpwgts = list(itertools.chain.from_iterable(tpwgts))
+
+    with _zero_numbering(options):
+        objval, part = _metis.part_graph(xadj, adjncy, nparts, vwgt, vsize, adjwgt,
+                                         tpwgts, ubvec, options, recursive)
+
+    parts = [[] for i in range(nparts)]
+    for u, i in zip(G, part):
+        parts[i].append(u)
+
+    return objval, parts
+
+
+@nx.utils.not_implemented_for('directed')
+@nx.utils.not_implemented_for('multigraph')
+@_convert_exceptions(nx.NetworkXError, (ValueError, TypeError, types.MetisError))
+def vertex_separator(G, weight='weight', options=None):
+    """Compute a vertex separator that bisects a graph. The algorithm aims to
+    minimize the sum of weights of vertices in the separator.
+
+    Parameters
+    ----------
+    G : NetworkX graph
+        A graph.
+
+    weight : object, optional
+        The data key used to determine the weight of each node. If None, each
+        node has unit weight. Default value: 'weight'.
+
+    options : MetisOptions, optional
+        METIS options. If None, the default options are used. Default value:
+        None.
+
+    Returns
+    -------
+    sep, part1, part2 : lists of nodes
+        The separator and the two parts of the bisection represented as lists.
+
+    Raises
+    ------
+    NetworkXError
+        If the parameters cannot be converted to valid METIS input format, or
+        METIS returns an error status.
+    """
+    if len(G) == 0:
+        return [], [], []
+
+    vwgt = [G.node[u].get(weight, 1) for u in G]
+    if all(w == 1 for w in vwgt):
+        vwgt = None
+
+    xadj, adjncy = _convert_graph(G)
+
+    with _zero_numbering(options):
+        part = _metis.compute_vertex_separator(xadj, adjncy, vwgt, options)[1]
+
+    groups = [[], [], []]
+    for u, i in zip(G, part):
+        groups[i].append(u)
+
+    return groups[2], groups[0], groups[1]
--- a/networkx/addons/metis/tests/test_metis.py
+++ b/networkx/addons/metis/tests/test_metis.py
@@ -1,0 +1,70 @@
+import itertools
+import nose.tools
+
+import _metis
+
+
+def make_cycle(n):
+    xadj = list(range(0, 2 * n + 1, 2))
+    adjncy = list(
+        itertools.chain.from_iterable(
+            zip(itertools.chain([n - 1], range(n - 1)),
+                itertools.chain(range(1, n), [0]))))
+    return xadj, adjncy
+
+
+class TestMetis(object):
+
+    def test_node_nd(self):
+        n = 16
+        xadj, adjncy = make_cycle(n)
+        perm, iperm = _metis.node_nd(xadj, adjncy)
+        nose.tools.assert_equal(set(perm), set(range(n)))
+        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
+        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
+                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
+        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+
+    def test_selfloops(self):
+        n = 16
+        xadj = list(range(0, 3 * n + 1, 3))
+        adjncy = list(
+            itertools.chain.from_iterable(
+                zip(itertools.chain([n - 1], range(n - 1)),
+                    range(n),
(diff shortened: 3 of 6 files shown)


OrkoHunter:
I had to rename `types.py` to avoid conflicts with the python built-in module `types`.

metis.py
```diff
@@ -0,0 +1,272 @@
...
+from decorators import convert_exceptions
+from _types import MetisError
+
+try:
+    from _metis import node_nd, part_graph, compute_vertex_separator
+    from _types import MetisOptions
+    __all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
+               'MetisOptions']
+except ImportError:
+    __all__ = []
```
ysitu: After a success installation, these imports should just work. Let's not hide any errors.

decorators.py
```diff
@@ -0,0 +1,53 @@
...
+            exc = e
+        gvars = globals()
+        lvars = locals()
+        lvars['convert_type'] = convert_type
+        if version_info[0] >= 3:
+            code = 'raise convert_type(exc).with_traceback(exc.__traceback__)'
+        else:
+            code = 'raise convert_type, (exc,), exc_info()[2]'
+        _exec(code, gvars, lvars)
+    return _convert_exceptions
```
ysitu: Add new line.

functions.py
```diff
@@ -0,0 +1,5 @@
+try:
+    from _metis import *
+    __all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
+except ImportError:
+    pass
```
ysitu: Is this file necessary?

metis.py
```diff
@@ -0,0 +1,272 @@
...
+    """
+    if len(G) == 0:
+        return []
+
+    vwgt = [G.node[u].get(weight, 1) for u in G]
+    if all(w == 1 for w in vwgt):
+        vwgt = None
+
+    if G.is_directed() or G.is_multigraph():
+        G = nx.Graph(G.edges_iter())
```
ysitu: `G.edges()`

functions.py
OrkoHunter: I think no. I guess it might be added to give output for the functions metis provide as `metis.functions.__all__`.
But that can be read from `metis.py` too. I'll remove it.

tests/test_metis.py
```diff
@@ -0,0 +1,74 @@
...
+    xadj = list(range(0, 2 * n + 1, 2))
+    adjncy = list(chain.from_iterable(zip(chain([n - 1], range(n - 1)),
+                                          chain(range(1, n), [0]))))
+    return xadj, adjncy
+
+
+class TestMetis:
+
+    @classmethod
+    def setupClass(cls):
```
ysitu: Remove this function.

tests/test_metis.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def make_cycle(n):
+    xadj = list(range(0, 2 * n + 1, 2))
+    adjncy = list(chain.from_iterable(zip(chain([n - 1], range(n - 1)),
+                                          chain(range(1, n), [0]))))
+    return xadj, adjncy
+
+
+class TestMetis:
```
ysitu: `class TestMetis(object)`:

_types.py
```diff
@@ -0,0 +1,190 @@
+from numbers import Number
+
+from enums import *
+try:
+    from _metis import set_default_options
+except ImportError:
```
ysitu: Remove `try... except`. `_metis` should just exist.

metis.py
```diff
@@ -0,0 +1,267 @@
...
+from sys import exc_info
+
+import networkx as nx
+from networkx.utils import not_implemented_for
+
+from decorators import convert_exceptions
+from _types import MetisError, MetisOptions
+from _metis import node_nd, part_graph, compute_vertex_separator
+__all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
+               'MetisOptions']
```
ysitu: Indentation

metis.py
ysitu: `edges_iter()` is being removed. Please fix.

decorators.py
```diff
@@ -0,0 +1,53 @@
...
+                raise
+            exc = e
+        gvars = globals()
+        lvars = locals()
+        lvars['convert_type'] = convert_type
+        if version_info[0] >= 3:
+            code = 'raise convert_type(exc).with_traceback(exc.__traceback__)'
+        else:
+            code = 'raise convert_type, (exc,), exc_info()[2]'
+        _exec(code, gvars, lvars)
```
ysitu: I have

```python
def _exec(code, gvars, lvars):
    exec(code, gvars, lvars)
```

in my repository. I forgot why I added this, but IIRC removing `_exec` broke something.

chebee7i: We'll definitely have some comments this time around? :)

OrkoHunter: A similar thing had happened with me, with `types.py`. Renaming it to `_types.py` worked for me but not the previous one because
```python
In [1]: import types
```
This always imported the built-in module `types` which I didn't know of.
```python
In [2]: types
Out[2]: <module 'types' from '/usr/lib/python2.7/types.pyc'>
```

OrkoHunter: @ysitu I added the `_exec` function. I might have missed it while taking from your repository.

chebee7i: Regarding `import types` failing, you should use relative imports if you want to avoid that---but a rename works just fine too.  `from . import types`.

Let's make sure we document why `_exec` is necessary.

OrkoHunter: > Let's make sure we document why _exec is necessary.

Yeah. We'll try it after we successfully build the library and find which `break` @ysitu is talking about.

> from . import types

This used to raise `Attempted relative import from non-package error`. Due to a namespace package, I guess we may not put `__init__.py` on the top.

chebee7i:
>> from . import types

> This used to raise `Attempted relative import from non-package error`. Due to a namespace package, I guess we may not put `__init__.py` on the top.

That's a really unappealing "feature" of namespace packages. It seems like any sufficiently complicated namespace package would want the option to do relative imports, rather than having to hardcode the package name in absolute imports.  The use case of a namespace package consisting of more than one module isn't discussed in [PEP 420](https://www.python.org/dev/peps/pep-0420/)

And, would this even work in Python 3? Python 3 does not allow implicit relative imports at all. Also, note that the way to create namespace packages differs in 3.x than it does in 2.x.

Though I haven't dug into this, it seems like a namespace package is potentially more complicated then doing the simple approach:

```
# networkx/metis/__init__.py

from __future__ import absolute_import

# this is an absolute import to .../site-packages/python2.7/networkx_metis
from networkx_metis import *
```

Then users use it as:   `import networkx.metis as metis`.

ysitu:
How about not using relative import at all? We know where this package will land after installation.

OrkoHunter:
@ysitu That is good. Maybe we can change all the imports before the final release of `networkx-metis`. As of now, it's under development, we should just keep it so.

chebee7i:
Yeah that's fine. Have you guys looked into how to do this in a 2/3 compatible way? Would we just use `pkgutil` or would we use setuptools?

chebee7i:
And I do realize this is minutiae compared to the larger compilation issues you are focusing on now. So feel free to ignore my comments until they become relevant.

OrkoHunter:
@chebee7i No, I'm reading things about it in particular and I think `setuptools` would be the way to go. However, we'll discuss about it.

_types.py
```diff
@@ -0,0 +1,187 @@
+from numbers import Number
+
+from enums import *
+from _metis import set_default_options
```
OrkoHunter: This seems to create a problem of cyclic import. `_metis` imports `_types` and here `_types` import `_metis` back.

ysitu: What about `import _metis` and use `_metis.set_default_options(self)` below?

ysitu: BTW, should we consider consistently using `import some_module` and `some_module.some_function` instead of `from some_module import some_function`?

OrkoHunter: This diff works.

```diff
diff --git a/_metis.pyx b/_metis.pyx
index 61cccd3..fc140c7 100644
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -4,7 +4,7 @@ from os import dup, dup2, fdopen
 from sys import stdout
 from tempfile import TemporaryFile
 from _api cimport *
-from ._types import MetisError
+import _types
 
 __all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
 
@@ -134,7 +134,7 @@ cdef void convert_graph(xadj, adjncy, idx_t *nvtxs_ptr, idx_t **_xadj_ptr,
 
 cdef void check_result(int result, msg) except *:
     if result != METIS_OK:
-        raise MetisError(msg)
+        raise _types.MetisError(msg)
 
 
 def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
diff --git a/_types.py b/_types.py
index 0f4339a..7001826 100644
--- a/_types.py
+++ b/_types.py
@@ -1,7 +1,6 @@
 from numbers import Number
 
-from enums import *
-from _metis import set_default_options
+import _metis
 
 __all__ = ['MetisOptions', 'MetisError']
 
@@ -9,7 +8,7 @@ __all__ = ['MetisOptions', 'MetisError']
 class MetisOptions(object):
 
     def __init__(self):
-        set_default_options(self)
+        _metis.set_default_options(self)
 
     @property
     def ptype(self):
diff --git a/metis.py b/metis.py
index 0dffe9b..3fa87be 100644
--- a/metis.py
+++ b/metis.py
@@ -14,11 +14,12 @@ from sys import exc_info
 import networkx as nx
 from networkx.utils import not_implemented_for
 
-from decorators import convert_exceptions
-from _types import MetisError, MetisOptions
-from _metis import node_nd, part_graph, compute_vertex_separator
+from enums import *
+import decorators
+import _types
+import _metis
 __all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
-           'MetisOptions']
+           '_types.MetisOptions']
 
 
 @contextmanager
@@ -46,7 +47,7 @@ def _convert_graph(G):
     return xadj, adjncy
 
 
-@convert_exceptions(nx.NetworkXError, (ValueError, TypeError, MetisError))
+@decorators.convert_exceptions(nx.NetworkXError, (ValueError, TypeError, _types.MetisError))
 def node_nested_dissection(G, weight='weight', options=None):
     """Compute a node ordering of a graph that reduces fill when the Laplacian
     matrix of the graph is LU factorized. The algorithm aims to minimize the
@@ -88,7 +89,7 @@ def node_nested_dissection(G, weight='weight', options=None):
     xadj, adjncy = _convert_graph(G)
 
     with _zero_numbering(options):
-        perm = node_nd(xadj, adjncy, vwgt, options)[0]
+        perm = _metis.node_nd(xadj, adjncy, vwgt, options)[0]
 
     nodes = list(G)
     perm = [nodes[i] for i in perm]
@@ -98,7 +99,7 @@ def node_nested_dissection(G, weight='weight', options=None):
 
 @not_implemented_for('directed')
 @not_implemented_for('multigraph')
-@convert_exceptions(nx.NetworkXError, (ValueError, TypeError, MetisError))
+@decorators.convert_exceptions(nx.NetworkXError, (ValueError, TypeError, _types.MetisError))
 def partition(G, nparts, node_weight='weight', node_size='size',
               edge_weight='weight', tpwgts=None, ubvec=None, options=None,
               recursive=False):
@@ -208,7 +209,7 @@ def partition(G, nparts, node_weight='weight', node_size='size',
         tpwgts = list(chain.from_iterable(tpwgts))
 
     with _zero_numbering(options):
-        objval, part = part_graph(xadj, adjncy, nparts, vwgt, vsize, adjwgt,
+        objval, part = _metis.part_graph(xadj, adjncy, nparts, vwgt, vsize, adjwgt,
                                   tpwgts, ubvec, options, recursive)
 
     parts = [[] for i in range(nparts)]
@@ -218,7 +219,7 @@ def partition(G, nparts, node_weight='weight', node_size='size',
     return objval, parts
 
 
-@convert_exceptions(nx.NetworkXError, (ValueError, TypeError, MetisError))
+@decorators.convert_exceptions(nx.NetworkXError, (ValueError, TypeError, _types.MetisError))
 def vertex_separator(G, weight='weight', options=None):
     """Compute a vertex separator that bisects a graph. The algorithm aims to
     minimize the sum of weights of vertices in the separator.
@@ -259,7 +260,7 @@ def vertex_separator(G, weight='weight', options=None):
     xadj, adjncy = _convert_graph(G)
 
     with _zero_numbering(options):
-        part = compute_vertex_separator(xadj, adjncy, vwgt, options)[1]
+        part = _metis.compute_vertex_separator(xadj, adjncy, vwgt, options)[1]
 
     groups = [[], [], []]
     for u, i in zip(G, part):
```

metis.py
```diff
@@ -0,0 +1,268 @@
...
+from sys import exc_info
+
+import networkx as nx
+from networkx.utils import not_implemented_for
+
+import decorators
+import _types
+import _metis
+__all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
+           '_types.MetisOptions']
```
ysitu: You can bring `MetisOptions` into scope by putting

```python
MetisOption = _types.MetisOption
```

in the global scope. This is a frequently used technique.

metis.py
```diff
@@ -0,0 +1,268 @@
...
+    finally:
+        if options:
+            options.numbering = numbering
+
+
+def _convert_graph(G):
+    """Convert a graph to the numbered adjacency list structure expected by
+    METIS.
+    """
+    index = dict(list(zip(G, list(range(len(G))))))
```
ysitu: Not sure why I did this, but

```python
index = dict(zip(G, range(len(G))))
```

makes better sense.

metis.py
```diff
@@ -0,0 +1,268 @@
...
+
+import decorators
+import _types
+import _metis
+__all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
+           '_types.MetisOptions']
+
+
+@contextmanager
+def _zero_numbering(options):
```
ysitu: Add a docstring

```python
"""Temporarily force zero-based numbering."""
```

metis.py
```diff
@@ -0,0 +1,268 @@
...
+        METIS returns an error status.
+    """
+    if len(G) == 0:
+        return []
+
+    vwgt = [G.node[u].get(weight, 1) for u in G]
+    if all(w == 1 for w in vwgt):
+        vwgt = None
+
+    if G.is_directed() or G.is_multigraph():
```
ysitu: Can we ignore `G.is_multigraph()`?

metis.py
OrkoHunter: Actually this was modified by `2to3`, you already had `index = dict(zip(G, range(len(G))))`.
`2to3` seems to make this change but it's not making sense. I'll change it.

metis.py
```diff
@@ -0,0 +1,268 @@
...
+    index = dict(list(zip(G, list(range(len(G))))))
+    xadj = [0]
+    adjncy = []
+    for u in G:
+        adjncy.extend(index[v] for v in G[u])
+        xadj.append(len(adjncy))
+    return xadj, adjncy
+
+
+@decorators.convert_exceptions(nx.NetworkXError, (ValueError, TypeError, _types.MetisError))
```
ysitu: Move `convert_exceptions` into this module and rename it `_convert_exceptions`.

metis.py
OrkoHunter: I don't think `partition()` is implemented for either directed graph or multi graph so, maybe we can ignore it here.

metis.py
OrkoHunter: Done.

Regarding the separate `_exec` function.
```py
def _exec(code, gvars, lvars):
    exec(code, gvars, lvars)
```
It seems that `exec` function can not be used in a function which has nested function in it with local variables. So, there's a way to use `exec` as 
```py
exec() in globals(), locals()
```
Or making a dummy function `_exec` and do the same job.
Should we go with `_exec` or use it as `exec() in globals(), locals()`?

metis.py
```diff
@@ -0,0 +1,320 @@
...
+    adjncy = []
+    for u in G:
+        adjncy.extend(index[v] for v in G[u])
+        xadj.append(len(adjncy))
+    return xadj, adjncy
+
+def _exec(code, gvars, lvars):
+    exec(code, gvars, lvars)
+
+def _convert_exceptions(convert_type, catch_types=None):
```
ysitu: Use two blank lines between top-level function definitions.

OrkoHunter: Added a line here and three other places.

Himanshu Mishra

metis.py
```diff
@@ -0,0 +1,323 @@
...
+    xadj = [0]
+    adjncy = []
+    for u in G:
+        adjncy.extend(index[v] for v in G[u])
+        xadj.append(len(adjncy))
+    return xadj, adjncy
+
+
+def _exec(code, gvars, lvars):
+    exec(code, gvars, lvars)
```
ysitu: Document why we need this function as requested by @chebee7i.

OrkoHunter: Yeah sure, but first please have a look at [this comment](https://github.com/networkx/networkx-metis/pull/3#discussion_r32720638). We don't actually need this function.

chebee7i:
Are you guys depending on `six`. It looks like it provides `reraise` which let's you do this without having to code up `_exec`.

http://stackoverflow.com/a/14505093

metis.py
ysitu: Is this compatible with Python 3?

ysitu:
@chebee7i That is actually a good suggestion. Using `six` in this case is certainly a big plus in code readability.

metis.py
ysitu: Actually I meant outright banning digraphs and multigraphs with `@not_implemented_for`.

OrkoHunter: Okay cool. I put `@not_implemented_for` for both graphs.

OrkoHunter:
@chebee7i Thanks for the suggestion. `six` seems to solve our problem in one another way.
The `exec` statement was replaced with an `exec(...)` function in Python 3. Replace:
```
exec code in globals, locals
```
With:
```
import six
six.exec_(code, globals, locals)
```
Source : https://github.com/ipython/ipython/wiki/IPEP-4:-Python-3-Compatibility#syntax-changes

Should I remove `_exec` and use it?

metis.py
```diff
@@ -0,0 +1,320 @@
...
+        perm = _metis.node_nd(xadj, adjncy, vwgt, options)[0]
+
+    nodes = list(G)
+    perm = [nodes[i] for i in perm]
+
+    return perm
+
+
+@nx.utils.decorators.not_implemented_for('directed')
+@nx.utils.decorators.not_implemented_for('multigraph')
```
SanketDG: Could be just:
```python
@nx.utils.not_implemented_for('directed')
@nx.utils.not_implemented_for('multigraph')
```
Or better, import once and use many times:
```python
from networkx.utils import not_implemented_for

@not_implemented_for('directed')
.....
```

OrkoHunter: I guess, for
```py
@nx.utils.not_implemented_for('directed')
```
this will be needed
```py
from networkx.utils import not_implemented_for
```
The later implementation was there earlier which seem to be very tempting too but for that we'd be going against the standard import style which suggests you do not import names from other modules into yours, rather import module and then use `module.name`.

However, this seems to be a better way. Declaring a local name
```
not_implemented_for = nx.utils.decorators.not_implemented_for
```
and using it at places.

SanketDG: > I guess, for
> 
> @nx.utils.not_implemented_for('directed')
> this will be needed
> 
> from networkx.utils import not_implemented_for

I think `networkx.utils` is imported in the top-level namespace. So you don't need the import statement.

> The later implementation was there earlier which seem to be very tempting too but for that we'd be going against the standard import style which suggests you do not import names from other modules into yours, rather import module and then use module.name.
> 
> However, this seems to be a better way. Declaring a local name
> 
> not_implemented_for = nx.utils.decorators.not_implemented_for
> and using it at places.

Yes, this is okay, but just:
```python
not_implemented_for = nx.utils.not_implemented_for
```
would suffice since `not_implemented_for` is imported in `networkx.utils`

OrkoHunter: Yes this is right. Thanks.

OrkoHunter: Changed it to `@nx.utils.not_implemented_for`. Probably this is more readable too.

chebee7i:
Doesn't reraise do exactly what you are intending with those lines `_convert_exceptions`? If I understand it correctly, you won't need to call `six.exec_`, `six.reraise` will handle that for you.

OrkoHunter:
So, if we are not using `six.exec_(code, globals, locals)`, then I'm a little unclear about the syntax with `six.raise()`. What would that be?

chebee7i:
I haven't really looked into it, but my first guess is that something like the following could work:

```python
   def _convert_exceptions(func, *args, **kwargs):
       try:
           return func(*args, **kwargs)
       except catch_types or () as e:
           exc = sys.exec_info()
       except Exception as e:
           if catch_types is not None:
               raise
           exc = sys.exec_info()
       six.reraise(convert_type, convert_type(exc[1]), exc[2])
   return _convert_exceptions
```

OrkoHunter:
@chebee7i  That seems to work just fine. Thanks!

chebee7i:
It imports/compiles or it passes all the tests? :)

OrkoHunter:
@chebee7i Yes it compiles. Tests are not passing anyway.

metis.py
```diff
@@ -0,0 +1,309 @@
...
+
+import contextlib
+import itertools
+import sys
+import decorator
+
+import networkx as nx
+
+import _types
+import _metis
```
ysitu: Shall we move to use absolute imports now?

OrkoHunter: I didn't get it. I though this was preferred.

ysitu: Sorry for being vague. I meant something like `from networkx.addons.metis import _types`.

OrkoHunter: Doing this might force a user to install `networkx-metis` into their system before using it. In the current way, I think the user is free to install as well as test locally without installing `networkx-metis` which I personally think, is a great advantage.
Do you think the current system might cause some problem ahead as we are pass with relative imports?

chebee7i: As far as I understand, you cannot do imports like `import _metis` in Python 3. All imports are read from sys.path and are treated as absolute (e.g. the local directory is not searched unless it is explicitly added). See here: https://www.python.org/dev/peps/pep-0328. This is precisely what we were discussing in another thread.

chebee7i: In other words....  unless you have a dot such as `from . import _metis` then it is considered an absolute import from sys.path. Since we can't have `__init__.py` files in the namespace package, we can't do relative imports anyway.

ysitu: I don't understand the worry about the user having to install `networkx-metis`. Installation is required anyhow---you need to build the Cython extensions.

OrkoHunter: Yeah this is right. We surely need to change them. Updating.

OrkoHunter: Updated `_types.py` and `metis.py` with the new import style.

OrkoHunter:
Updated with the required directory structure.
```
networkx/
     __init__.py
    addons/
        __init.py
        metis/
            metis.py
            _types.py
            enums.py
            tests/
                test_metis.py
```

chebee7i:
Is it still necessary to have it named _types? Absolute imports the default and in 2.7 you should be using `from __future__ import absolute_import`.

OrkoHunter:
@chebee7i No, we can now keep both the names as well. Do you prefer changing it back to `types.py`?

chebee7i:
I don't know what the reason for having an underscore should be, now that there is no conflict. So it seems like it should go back to `types.py`.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,187 @@
+import numbers
+
+import enums
+from networkx.addons.metis import _metis
+
+__all__ = ['MetisOptions', 'MetisError']
+
+
+class MetisOptions(object):
```
ysitu: Add docstrings to class and properties.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,187 @@
...
+        names = ['ptype', 'objtype', 'ctype', 'iptype', 'rtype', 'ncuts',
+                 'nseps', 'numbering', 'niter', 'seed', 'minconn', 'no2hop',
+                 'contig', 'compress', 'ccorder', 'pfactor', 'ufactor']
+        return '{0}({1})'.format(
+            self.__class__.__name__,
+            ', '.join('{0}={1}'.format(name, repr(getattr(self, name)))
+                      for name in names))
+
+
+class MetisError(Exception):
```
ysitu: Add docstrings.

networkx/addons/metis/enums.py
```diff
@@ -0,0 +1,118 @@
...
+class MetisPType(enum.IntEnum):
+    """Partitioning method.
+
+        ==== =================================
+        rb   Multilevel recursive bisectioning
+        kway Multilevel `k`-way partitioning
+        ==== =================================
+    """
+
+    default, rb, kway = list(range(-1, 2))
```
ysitu: `list()` is unnecessary.

networkx/addons/metis/metis.py
```diff
@@ -0,0 +1,309 @@
...
+# All rights reserved
+
+"""
+Wrappers of METIS graph partitioning functions.
+"""
+
+import contextlib
+import itertools
+import sys
+import decorator
```
ysitu: Order these alphabetically.

networkx/addons/metis/metis.py
```diff
@@ -0,0 +1,309 @@
...
+
+import networkx as nx
+
+from networkx.addons.metis import _types
+from networkx.addons.metis import _metis
```
ysitu: Order these too.

ysitu:
Please address the comments about docstrings so that we can merge this PR and unblock some others. For the most part, this should involve only duplicating some text from `enums.py` into `types.py`.

OrkoHunter:
@ysitu On it.

OrkoHunter:
I should rename `_types` back to `types` right?

OrkoHunter:
Docstrings added.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class MetisOptions(object):
+    """Assign default values to a MetisOptions object."""
+    def __init__(self):
+        _metis.set_default_options(self)
+
+    @property
+    def ptype(self):
+        """Types of Partitioning method"""
```
ysitu: I think that it is more useful to the user when they say `help(MetisOptions.ptype)` they get a list of valid enums instead of just two words. You should copy the entire docstring block here.

OrkoHunter: That's nice.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,206 @@
+import numbers
+
+import enums
+from networkx.addons.metis import _metis
+
+__all__ = ['MetisOptions', 'MetisError']
+
+
+class MetisOptions(object):
+    """Assign default values to a MetisOptions object."""
```
ysitu: Should be `Options controlling behaviors of METIS algorithms.`. Also add a blank line before `def __init__()`.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self):
+        _metis.set_default_options(self)
```
ysitu: This also needs a docstring. `Initializes a MetisOptions object.` should be good enough.

BTW, it is probably useful to enable this syntax:

```python
options = MetisOptions(niter=100)
```

so that you can avoid the verbosity in

```python
options = MetisOptions()
options.niter = 100
```

An extra `**kwargs` parameter should do the trick. You need to extend the docstring for that, though.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,272 @@
...
+import enums
+from networkx.addons.metis import _metis
+
+__all__ = ['MetisOptions', 'MetisError']
+
+
+class MetisOptions(object):
+    """Options controlling behaviors of METIS algorithms"""
+
+    def __init__(self, **kwargs):
```
OrkoHunter: Updated with `**kwargs` option and docstring with one example.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,272 @@
...
+        """Initializes a MetisOptions object. Values can be provided
+        for some parameters as arguments.
+
+        Example
+        -------
+        >>> options = MetisOptions(ncuts=2, niter=100)
+
+        """
+        _metis.set_default_options(self)
```
ysitu: You need to do something with `kwargs`, right?

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,272 @@
...
+    def __init__(self, **kwargs):
+        """Initializes a MetisOptions object. Values can be provided
+        for some parameters as arguments.
+
+        Example
+        -------
+        >>> options = MetisOptions(ncuts=2, niter=100)
+
+        """
+        _metis.set_default_options(self, **kwargs)
```
OrkoHunter: Will this do?

ysitu: No. You cannot just try things randomly. You need to think about what things would work and why. In this case, when the user says `MetisOptions(niter=100)`, you want to achieve the effect of `options.niter = 100`. You need to find out where to extract `niter` and `100` and how to set them on `options`.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,275 @@
...
+        Example
+        -------
+        >>> options = MetisOptions(ncuts=2, niter=100)
+
+        """
+        _metis.set_default_options(self)
+        for key, value in kwargs.items():
+            if (value != None):
+                setattr(self, '_' + key, value)
+
```
OrkoHunter: This one should go. It works.

ysitu: You should not add an underscore, otherwise the validation checks will be bypassed.

networkx/addons/metis/_types.py
```diff
@@ -0,0 +1,275 @@
...
+        for some parameters as arguments.
+
+        Example
+        -------
+        >>> options = MetisOptions(ncuts=2, niter=100)
+
+        """
+        _metis.set_default_options(self)
+        for key, value in kwargs.items():
+            if (value != None):
```
ysitu: Remove parentheses. Always use `is not None` instead of `!= None`.

networkx/addons/metis/tests/test_metis.py
```diff
@@ -0,0 +1,70 @@
...
+        adjncy = list(
+            itertools.chain.from_iterable(
+                zip(itertools.chain([n - 1], range(n - 1)),
+                    range(n),
+                    itertools.chain(range(1, n), [0]))))
+        perm, iperm = _metis.node_nd(xadj, adjncy)
+        nose.tools.assert_equal(set(perm), set(range(n)))
+        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
+        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
+                      (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
```
ysitu: One extra space of indentation.

networkx/addons/metis/tests/test_metis.py
```diff
@@ -0,0 +1,70 @@
...
+class TestMetis(object):
+
+    def test_node_nd(self):
+        n = 16
+        xadj, adjncy = make_cycle(n)
+        perm, iperm = _metis.node_nd(xadj, adjncy)
+        nose.tools.assert_equal(set(perm), set(range(n)))
+        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
+        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
+                      (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
```
ysitu: One extra space of indentation.

ysitu:
> I should rename _types back to types right?

With absolute imports now it makes sense to rename.

ysitu:
I will merge this after the last three comments are addressed.

OrkoHunter:
> I will merge this after the last three comments are addressed.

Shouldn't I add the https://github.com/networkx/networkx-metis/pull/14#issuecomment-115105652 about error raised in setting the value of `niter` in this PR?

ysitu:
You can keep the new test in #14.

OrkoHunter:
Indentations corrected. `_types` renamed back to `types` and edits made in `metis.py`

OrkoHunter:
There's a [line](https://github.com/networkx/networkx-metis/blob/master/_metis.pyx#L7) in `_metis.pyx` where `_types` is being imported. How do you think should I make that change? With a commit in #14 or a new PR?

ysitu:
It is okay to piggyback that on #14.

OrkoHunter:
Okay. Thanks for merging!

PR #6, opened 2015-05-27
[WIP] Add license and NOTICE

OrkoHunter:
LGPL's basic requirement is to separate the LGPL-licensed library and our own product well enough. That should allow users to supply their own version of the library instead of the one we've shipped with our software (with the bugs fixed, for instance). 

Note, however, that mere separation is not enough, though required. We should provide our users a documented way to replace a library with their version (i.e., how to upload firmware, or to recompile a Python wrapper for an LGPL C library).

The second notable clause is attribution requirement. This should help to promote the name of the original developer of the library. If our license is Apache, this would be in a `NOTICE` file. We should list the name of the LGPL work that we used.

References:
* http://programmers.stackexchange.com/questions/86142/what-exactly-do-i-need-to-do-if-i-use-a-lgpl-licenced-library
* http://answers.google.com/answers/threadview/id/439136.html
* http://stackoverflow.com/questions/1383919/does-the-lgpl-ensure-that-programs-that-rely-on-my-code-admit-it

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,13 @@
+Copyright 2015 [?Name of Copyright owner?]
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


LICENSE.txt
```diff
@@ -0,0 +1,13 @@
+Copyright 2015 [?Name of Copyright owner?]
```
OrkoHunter: We will have to decide whether the owners are going to be NetworkX owners, or `NetworkX Developers` or the add-on creator.

OrkoHunter:
I guess this should be closed now. I'll send a different PR for the license.

PR #10, opened 2015-06-10
Fix import style

OrkoHunter:
With this change and updated #3, `networkx-metis` works fine on python 2. But has errors with python 3.
```python
Python 3.4.0 (default, Apr 11 2014, 13:05:11) 
Type "copyright", "credits" or "license" for more information.

In [1]: import metis
---------------------------------------------------------------------------
ImportError: Traceback (most recent call last)
<ipython-input-1-f3c558d985ea> in <module>()
----> 1 import metis

/metis.py in <module>()
     16 
     17 from decorators import convert_exceptions
---> 18 from _types import MetisError, MetisOptions
     19 from _metis import node_nd, part_graph, compute_vertex_separator
     20 __all__ = ['node_nested_dissection', 'partition', 'vertex_separator',

/_types.py in <module>()
      2 
      3 from enums import *
----> 4 from _metis import set_default_options
      5 
      6 __all__ = ['MetisOptions', 'MetisError']

ImportError: _metis.so: undefined symbol: _Py_ZeroStruct
```

### DIFF
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -1,101 +1,102 @@
-from cpython.mem cimport PyMem_Malloc, PyMem_Free
-from contextlib import contextmanager
-from os import dup, dup2, fdopen
-from sys import stdout
-from tempfile import TemporaryFile
-from _api cimport *
-from ._types import MetisError
+cimport cpython.mem
+import contextlib
+import os
+import sys
+import tempfile
+
+from networkx.addons.metis import _types
+cimport _api
 
 __all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
 
 
-@contextmanager
+@contextlib.contextmanager
 def redirect(source, target):
     """Temporarily redirect operations on file ``source`` to file ``target``.
     """
     source.flush()
     fd = source.fileno()
-    with fdopen(dup(fd), source.mode) as source2:
-        dup2(target.fileno(), fd)
+    with os.fdopen(os.dup(fd), source.mode) as source2:
+        os.dup2(target.fileno(), fd)
         try:
             yield
         finally:
             source.flush()
-            dup2(source2.fileno(), fd)
+            os.dup2(source2.fileno(), fd)
 
 
-cdef void* checked_malloc(idx_t size) except NULL:
-    cdef void* ptr = PyMem_Malloc(size)
+cdef void* checked_malloc(_api.idx_t size) except NULL:
+    cdef void* ptr = cpython.mem.PyMem_Malloc(size)
     if ptr == NULL:
         raise MemoryError()
     return ptr
 
 
-cdef idx_t* convert_idx_array(array) except NULL:
+cdef _api.idx_t* convert_idx_array(array) except NULL:
     """Convert a list of ints to a C array of idx_t's.
     """
-    cdef idx_t size = len(array)
-    cdef idx_t *_array = <idx_t*> checked_malloc(sizeof(idx_t) * size)
-    cdef idx_t i
+    cdef _api.idx_t size = len(array)
+    cdef _api.idx_t *_array = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * size)
+    cdef _api.idx_t i
     try:
         for i from 0 <= i < size:
             _array[i] = array[i]
     except:
-        PyMem_Free(_array)
+        cpython.mem.PyMem_Free(_array)
         raise
     return _array
 
 
-cdef real_t* convert_real_array(array) except NULL:
+cdef _api.real_t* convert_real_array(array) except NULL:
     """Convert a list of floats to a C array of real_t's.
     """
-    cdef idx_t size = len(array)
-    cdef real_t *_array = <real_t*> checked_malloc(sizeof(real_t) * size)
-    cdef idx_t i
+    cdef _api.idx_t size = len(array)
+    cdef _api.real_t *_array = <_api.real_t*> checked_malloc(sizeof(_api.real_t) * size)
+    cdef _api.idx_t i
     try:
         for i from 0 <= i < size:
             _array[i] = array[i]
     except:
-        PyMem_Free(_array)
+        cpython.mem.PyMem_Free(_array)
         raise
     return _array
 
 
-cdef convert_options(options, idx_t *_options):
+cdef convert_options(options, _api.idx_t *_options):
     """Convert a MetisOptions object to a C array.
     """
-    METIS_SetDefaultOptions(_options)
+    _api.METIS_SetDefaultOptions(_options)
     if options is None:
         return
 
-    _options[<idx_t> METIS_OPTION_PTYPE]     = options.ptype
-    _options[<idx_t> METIS_OPTION_OBJTYPE]   = options.objtype
-    _options[<idx_t> METIS_OPTION_CTYPE]     = options.ctype
-    _options[<idx_t> METIS_OPTION_IPTYPE]    = options.iptype
-    _options[<idx_t> METIS_OPTION_RTYPE]     = options.rtype
-    _options[<idx_t> METIS_OPTION_NCUTS]     = options.ncuts
-    _options[<idx_t> METIS_OPTION_NSEPS]     = options.nseps
-    _options[<idx_t> METIS_OPTION_NUMBERING] = options.numbering
-    _options[<idx_t> METIS_OPTION_NITER]     = options.niter
-    _options[<idx_t> METIS_OPTION_SEED]      = options.seed
-    _options[<idx_t> METIS_OPTION_MINCONN]   = options.minconn
-    _options[<idx_t> METIS_OPTION_NO2HOP]    = options.no2hop
-    _options[<idx_t> METIS_OPTION_CONTIG]    = options.contig
-    _options[<idx_t> METIS_OPTION_COMPRESS]  = options.compress
-    _options[<idx_t> METIS_OPTION_CCORDER]   = options.ccorder
-    _options[<idx_t> METIS_OPTION_PFACTOR]   = options.pfactor
-    _options[<idx_t> METIS_OPTION_UFACTOR]   = options.ufactor
-    _options[<idx_t> METIS_OPTION_DBGLVL]    = options.dbglvl
+    _options[<_api.idx_t> _api.METIS_OPTION_PTYPE]     = options.ptype
+    _options[<_api.idx_t> _api.METIS_OPTION_OBJTYPE]   = options.objtype
+    _options[<_api.idx_t> _api.METIS_OPTION_CTYPE]     = options.ctype
+    _options[<_api.idx_t> _api.METIS_OPTION_IPTYPE]    = options.iptype
+    _options[<_api.idx_t> _api.METIS_OPTION_RTYPE]     = options.rtype
+    _options[<_api.idx_t> _api.METIS_OPTION_NCUTS]     = options.ncuts
+    _options[<_api.idx_t> _api.METIS_OPTION_NSEPS]     = options.nseps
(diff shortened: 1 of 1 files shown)


_metis.pyx
```diff
@@ -134,6 +134,7 @@ cdef void convert_graph(xadj, adjncy, idx_t *nvtxs_ptr, idx_t **_xadj_ptr,
 
 cdef void check_result(int result, msg) except *:
     if result != METIS_OK:
+        from _types import MetisError
```
ysitu: This only hides the import problem until `check_result` fails. You probably need a top-level `__init__.py` instead.

OrkoHunter:
Regarding `ImportError: _metis.so: undefined symbol: _Py_ZeroStruct`, after some research I came to know that `_Py_ZeroStruct` symbol is not defined over Python 3. The evidence of this result can also be checked with a `grep` in `/usr/bin/` which gives positive result for the binary file `python2.7` but no result for other versions of python.

_metis.pyx
OrkoHunter: Explicitly I was trying to remove the `.` from `._types` but it seems both won't work here. And we can be sure that the problem lies only in the import system. Top level `__init__.py` is not doing the job either.

OrkoHunter: Should we go on with not using relative import at all as you suggested? Replacing
`from ._types import MetisError`
with
`from networkx.addons.metis.types import MetisError`

OrkoHunter: @dschult @chebee7i 
@ysitu might be away for few days. So more eyes would be helpful on it.

chebee7i: A top level init file is not allowed in namespace packages. @OrkoHunter if I wanted to try building this, should I checkout this branch? Do I need to merge another in with it?

OrkoHunter: @chebee7i  To try building, I think you need to fetch `sort-fix` and then merge it with `wrapper` and `setup`. This PR is just a minor change which needs more to be done.

OrkoHunter:
Updated. we should consider using `import modules; module.item()` instead of `from module import item`

_metis.pyx
```diff
@@ -4,7 +4,7 @@ from os import dup, dup2, fdopen
 from sys import stdout
 from tempfile import TemporaryFile
 from _api cimport *
-from ._types import MetisError
+import _types
```
chebee7i: Why does this work at all in Python 3? I was under the [impression](http://python3porting.com/problems.html#relative-import-problems) that this needed to be `from . import _types` (which is not allowed in a namespace package) or `import networkx.addons.metis._types as _types`.

OrkoHunter: `import networkx.addons.metis._types as _types` is doable. Should we go with that? However currently, it doesn't work in python3 as of [this comment](https://github.com/networkx/networkx-metis/pull/10#issuecomment-111456413).

ysitu: To test under Python 3 you need to run `python3 setup.py build_ext`.

_metis.pyx
```diff
@@ -22,21 +23,22 @@ def redirect(source, target):
         finally:
             source.flush()
             dup2(source2.fileno(), fd)
+            target.seek(0)
```
ysitu: Why add this back?

OrkoHunter: By mistake. Sorry, I'm just removing.

_metis.pyx
```diff
@@ -3,8 +3,9 @@ from contextlib import contextmanager
 from os import dup, dup2, fdopen
 from sys import stdout
 from tempfile import TemporaryFile
```
ysitu: What about these imports?

OrkoHunter: Ah! Too centered about `_api` so missed them. Updating soon.

OrkoHunter: Updated now!

OrkoHunter:
Updated `_metis.pyx` with
```
from networkx.addons.metis import _types
from networkx.addons.metis cimport _api
```

OrkoHunter:
This diff works for `StringIO instance has no attribute 'fileno'` Now the tests are passing. 

```diff
diff --git a/_metis.pyx b/_metis.pyx
index e13d742..b956af3 100644
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -264,7 +264,7 @@ def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
         _recursive = bool(recursive)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect(sys.__stdout__, tmp), nogil:
                 if _recursive:
                     result = _api.METIS_PartGraphRecursive(
                         &nvtxs, &ncon, _xadj, _adjncy, _vwgt, _vsize, _adjwgt,
@@ -274,7 +274,7 @@ def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
                         &nvtxs, &ncon, _xadj, _adjncy, _vwgt, _vsize, _adjwgt,
                         &_nparts, _tpwgts, _ubvec, _options, &objval, _part)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = unicode(tmp.read(), sys.__stdout__.encoding)
 
         check_result(result, msg)
 
@@ -354,11 +354,11 @@ def node_nd(xadj, adjncy, vwgt=None, options=None):
         _iperm = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * nvtxs)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect(sys.__stdout__, tmp), nogil:
                 result = _api.METIS_NodeND(&nvtxs, _xadj, _adjncy, _vwgt, _options,
                                       _perm, _iperm)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = unicode(tmp.read(), sys.__stdout__.encoding)
 
         check_result(result, msg)
 
@@ -425,11 +425,11 @@ def compute_vertex_separator(xadj, adjncy, vwgt=None, options=None):
         _part = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * nvtxs)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect(sys.__stdout__, tmp), nogil:
                 result = _api.METIS_ComputeVertexSeparator(
                     &nvtxs, _xadj, _adjncy, _vwgt, _options, &sepsize, _part)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = unicode(tmp.read(), sys.__stdout__.encoding)
 
         check_result(result, msg)
 
```

chebee7i:
Can you explain why you are reassigning `sys.__stdout__` rather than `sys.stdout`?  The double-underscored stdout is supposed to be a reference to the one true stdout.  When people reassign `sys.stdout`, they revert to the double-underscored version. https://docs.python.org/2/library/sys.html#sys.__stdout__

OrkoHunter:
I didn't read much about it but this was a fix for the failing test, I
found on stackoverflow. I'll look for a possible explanation.

debugger22:
You might want to read this https://docs.python.org/3.3/library/sys.html#sys.__stdout__

OrkoHunter:
It seems that the `nosetests` is the culprit. They probably replace `sys.stdout` with a `StringIO` object so that it can capture the output easily. But maybe they do not interfere with `sys.__stdout__`. I realize very well about the similarities of the two, and both works while importing/compiling but `nosetests` are making things worse.

Source : http://stackoverflow.com/questions/30961173/sys-stdout-works-but-sys-stdout-does-not#comment49954775_30961173

OrkoHunter:
There's more!
```
$ nosetests -s
....
Ran 4 tests in 0.009s

OK

```
works with `sys.stdout`.

chebee7i:
The other option mentioned on that SO post, seems more robust to me than what is currently implemented (since it doesn't assume that stdout has an fd).

OrkoHunter:
Or should we go with `nosetests -s` on Travis without any changes?

chebee7i:
The `nosetests -s` still requires that *other* code has not replaced `sys.stdout` with a `StringIO` instance. So I don't think it is a general solution.

ysitu:
The reason why we need to redirect `stdout` is that `METIS` outputs error messages to `stdout`. We need to remember that METIS is a C library. It knows nothing about Python's `sys.stdout` (or `sys.__stdout__`) object. Therefore, what is important is that the file descriptor of `stdout` in C, which should be `1` on common operating systems. So I think that it would not be too bad that we use

```python
fd = sys.__stdout__.fileno() if sys.__stdout__ else 1
```

or even

```python
fd = 1
```

Separately from obtaining the file descriptor of `stdout`, [this thread](http://stackoverflow.com/questions/4675728/redirect-stdout-to-a-file-in-python) brings up a very good problem of reliably flushing `stdout`. We need to handle that.

chebee7i:
Got it. That makes the implementation a bit more specific than I thought---namely, it is specifically targeting stdout generated from C functions rather than Python programs. Makes sense.

_metis.pyx
```diff
@@ -1,101 +1,102 @@
-from cpython.mem cimport PyMem_Malloc, PyMem_Free
-from contextlib import contextmanager
-from os import dup, dup2, fdopen
-from sys import stdout
-from tempfile import TemporaryFile
-from _api cimport *
-from ._types import MetisError
+cimport cpython.mem
```
ysitu: Just a comment. When the import name gets verbose, you can do

```cython
from cpython cimport mem
```

instead.

ysitu:
> Separately from obtaining the file descriptor of stdout, this thread brings up a very good problem of reliably flushing stdout. We need to handle that.

BTW, we are dealing with this in Cython. So we do have (somewhat) direct access to C facilities. In this case, `cimport`ing `fflush` and `stdout` from `libc.stdio` and `fflush(stdout)` should suffice.

_metis.pyx
OrkoHunter: Thanks. I'll change it.
Btw., with updated directory structure, `python setup.py build_ext --inplace` builds `_metis.so` in the top level directory instead of the desired `networkx/metis/addons`. What should be done for it to be created in the desired place?

ysitu: Whether you should change this particular import is your call. It was a just reminder that you can void

```python
import a.very.long.name.that.is.longer.than.long

a.very.long.name.that.is.longer.than.long.enough()
```

As for the `--inplace` problem, I think that it is about time that you create an automated script that builds, installs and tests the library.

chebee7i: I believe where the .so file is placed is dictated by the `setup.py`:

```
ext_modules = cythonize([Extension('_metis', ['*.pyx'],
                                   include_dirs=['src/GKlib', 'src/libmetis'],
                                   libraries=['metis', 'gklib'])])
```

If you change `_metis` to `networkx/metis/addons/_metis`, you should get what you want. But you'll need to verify that it works how you want it.

OrkoHunter:
I guess this 66d4f92f375a90b054e0de10467e44023afbb9cd is pretty much okay to do. `sys.stdout.encoding` also needed to be changed.

_metis.pyx
OrkoHunter: @chebee7i Yes this is right. Yes, `_metis` in pretty much needed where `metis.py` would fall.

ysitu:
Can we separate the `stdout` changes to another PR?

OrkoHunter:
Seems good to create new PR for last commit.

OrkoHunter:
Will send a PR shortly with the changes related to `stdout`. Last commit removed from this PR.

OrkoHunter:
@ysitu Thanks!

PR #14, opened 2015-06-23
Fix for redirecting of stdout problem

OrkoHunter:
Fixes errors concerning StringIO instance of `sys.stdout`.

@ysitu This line
```
    msg = unicode(tmp.read(), sys.stdout.encoding)
```
needs to be changed too.
```
AttributeError: StringIO instance has no attribute 'encoding'
```

### DIFF
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -1,28 +1,32 @@
+from libc cimport stdio
 cimport cpython.mem
+cimport _api
+
 import contextlib
 import os
 import sys
 import tempfile
 
 from networkx.addons.metis import _types
-cimport _api
 
 __all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
 
 
+_STDOUT_FILENO = 1 #File descriptor for stdout
+
+
 @contextlib.contextmanager
-def redirect(source, target):
-    """Temporarily redirect operations on file ``source`` to file ``target``.
+def redirect_stdout(target):
+    """Temporarily redirect operations on ``stdout`` to file ``target``.
     """
-    source.flush()
-    fd = source.fileno()
-    with os.fdopen(os.dup(fd), source.mode) as source2:
-        os.dup2(target.fileno(), fd)
+    stdio.fflush(stdio.stdout)
+    with os.fdopen(os.dup(_STDOUT_FILENO), 'w') as stdout_backup:
+        os.dup2(target.fileno(), _STDOUT_FILENO)
         try:
             yield
         finally:
-            source.flush()
-            os.dup2(source2.fileno(), fd)
+            stdio.fflush(stdio.stdout)
+            os.dup2(stdout_backup.fileno(), _STDOUT_FILENO)
 
 
 cdef void* checked_malloc(_api.idx_t size) except NULL:
@@ -264,7 +268,7 @@ def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
         _recursive = bool(recursive)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect_stdout(tmp), nogil:
                 if _recursive:
                     result = _api.METIS_PartGraphRecursive(
                         &nvtxs, &ncon, _xadj, _adjncy, _vwgt, _vsize, _adjwgt,
@@ -274,7 +278,7 @@ def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
                         &nvtxs, &ncon, _xadj, _adjncy, _vwgt, _vsize, _adjwgt,
                         &_nparts, _tpwgts, _ubvec, _options, &objval, _part)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = tmp.read().decode('ascii')
 
         check_result(result, msg)
 
@@ -354,11 +358,11 @@ def node_nd(xadj, adjncy, vwgt=None, options=None):
         _iperm = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * nvtxs)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect_stdout(tmp), nogil:
                 result = _api.METIS_NodeND(&nvtxs, _xadj, _adjncy, _vwgt, _options,
-                                      _perm, _iperm)
+                                           _perm, _iperm)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = tmp.read().decode('ascii')
 
         check_result(result, msg)
 
@@ -425,11 +429,11 @@ def compute_vertex_separator(xadj, adjncy, vwgt=None, options=None):
         _part = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * nvtxs)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect_stdout(tmp), nogil:
                 result = _api.METIS_ComputeVertexSeparator(
                     &nvtxs, _xadj, _adjncy, _vwgt, _options, &sepsize, _part)
             tmp.seek(0)
-            msg = unicode(tmp.read(), sys.stdout.encoding)
+            msg = tmp.read().decode('ascii')
 
         check_result(result, msg)
 
--- a/_metis.pyx
+++ b/_metis.pyx
@@ -7,7 +7,7 @@ import os
 import sys
 import tempfile
 
-from networkx.addons.metis import _types
+from networkx.addons.metis import types
 
 __all__ = ['part_graph', 'node_nd', 'compute_vertex_separator']
 
@@ -139,7 +139,7 @@ cdef void convert_graph(xadj, adjncy, _api.idx_t *nvtxs_ptr, _api.idx_t **_xadj_
 
 cdef void check_result(int result, msg) except *:
     if result != _api.METIS_OK:
-        raise _types.MetisError(msg)
+        raise types.MetisError(msg)
 
 
 def part_graph(xadj, adjncy, nparts, vwgt=None, vsize=None, adjwgt=None,
--- a/networkx/addons/metis/tests/test_metis.py
+++ b/networkx/addons/metis/tests/test_metis.py
@@ -68,3 +68,12 @@ class TestMetis(object):
             sorted(map(sorted,
                        [[(sep[0] + i) % n for i in range(1, n // 2)],
                         [(sep[1] + i) % n for i in range(1, n // 2)]])))
+
+    def test_MetisOptions(self):
+        n = 16
+        xadj, adjncy = make_cycle(n)
+        options = types.MetisOptions(niter=-2)
+        nose.tools.assert_raises_regexp(types.MetisError,
+                                        'Input Error: Incorrect niter.',
+                                        _metis.part_graph, xadj, adjncy, 2,
+                                        options=options)


_metis.pyx
```diff
@@ -6,22 +6,26 @@ import tempfile
 
 from networkx.addons.metis import _types
 cimport _api
+from libc cimport stdio
```
ysitu: Organize the imports a bit?

_metis.pyx
```diff
@@ -6,22 +6,26 @@ import tempfile
...
 
 
+_STDOUT_FILENO = 1 #File descriptor for stdout
+
+
 @contextlib.contextmanager
-def redirect(source, target):
-    """Temporarily redirect operations on file ``source`` to file ``target``.
+def redirect_stdout(tmp):
+    """Temporarily redirect operations on ``stdout`` to file ``tmp``.
```
ysitu: Keep the `target` name.

_metis.pyx
```diff
@@ -6,22 +6,26 @@ import tempfile
...
     """
-    source.flush()
-    fd = source.fileno()
-    with os.fdopen(os.dup(fd), source.mode) as source2:
-        os.dup2(target.fileno(), fd)
+    stdio.fflush(stdio.stdout)
+    fd = _STDOUT_FILENO
```
ysitu: Inline `_STDOUT_FILENO`.

_metis.pyx
```diff
@@ -6,22 +6,26 @@ import tempfile
...
+    with os.fdopen(os.dup(fd), 'w') as source2:
```
ysitu: Is it sufficient to merely save the value `dup(fd)` without `fdopen`ing?

_metis.pyx
OrkoHunter: Ascending order by length here?

_metis.pyx
OrkoHunter: Didn't get this one. What is meant by *inlining* here?

ysitu:
> msg = unicode(tmp.read(), sys.stdout.encoding)

How about `str(tmp.read(), 'ascii')`?

_metis.pyx
ysitu: Use `_STDOUT_FILENO` directly without assigning to `fd`.

_metis.pyx
ysitu: Maybe move `cimport`s to the top, follow Python imports.

OrkoHunter:
> How about str(tmp.read(), 'ascii')?

`str()` does not take two arguments. Can't we use `decode()` or `encode()` functions here?

ysitu:
`decode()` is the right choice.

_metis.pyx
```diff
@@ -1,28 +1,32 @@
...
 cimport cpython.mem
+cimport _api
+
 import contextlib
 import os
 import sys
 import tempfile
 
 from networkx.addons.metis import _types
-cimport _api
```
OrkoHunter: How about this? `cimports` organized according to system and local modules, and so are `imports` following them.

OrkoHunter:
And so, it works! :)

_metis.pyx
```diff
@@ -1,28 +1,32 @@
...
-        os.dup2(target.fileno(), fd)
+    stdio.fflush(stdio.stdout)
+    with os.fdopen(os.dup(_STDOUT_FILENO), 'w') as source2:
+        os.dup2(target.fileno(), _STDOUT_FILENO)
         try:
             yield
         finally:
-            source.flush()
-            os.dup2(source2.fileno(), fd)
+            stdio.fflush(stdio.stdout)
```
SanketDG: Since we are just using two functions, couldn't we import just `fflush` and `stdio` from the top of the file,
```python
from libc.stdio cimport fflush, stdout
```
And then:
```python
fflush(stdout)
```
Its just about code style, but it looks much cleaner.

OrkoHunter: As I feel, using bare `stdout` and `fflush` might become confusing sometime, with python's own `stdout` and `flush`. `stdio` seems to be adding a little bit more information. And that is why most of the times, using `module.func` is appreciated.

OrkoHunter:
We need to update the directory structure of the existing files in the `master`. Should I add one commit here and change the name of PR or make a separate one after this goes in?

ysitu:
Make a separate PR.

OrkoHunter:
That would be possible after this goes in, to avoid merge conflicts.

ysitu:
Are you able to test this PR?

OrkoHunter:
@ysitu Yes, and it works. I have a local setup which does not uses the namespace packing for installation. So, I test the changes there manually.

Regarding the automated script and Travis CI, that'd be working as soon as the namespace installation starts working for `networkx-metis`.

ysitu:
Do you mean that you have a test case that forces METIS to generate an error message, and you can verify that the error message is correctly captured?

OrkoHunter:
I fetched all the files in the current master plus files currently split into three PRs at one place, in one test directory and then I run 
```
$ python setup.py build_clib
```
 followed by 
```
$ python setup.py build_ext --inplace
```
And for this particular PR, I run `nosetests` which verify the tests written under `tests/test_metis.py` and then I manually try partitioning a graph by importing `networkx`.

In the current situation
```sh
$ nosetests                                               [master] 10:52:53
....
Ran 4 tests in 0.096s

OK
```

ysitu:
Those test cases do not really exercise the code paths that deal with exceptions.

_metis.pyx
```diff
@@ -354,11 +358,11 @@ def node_nd(xadj, adjncy, vwgt=None, options=None):
         _iperm = <_api.idx_t*> checked_malloc(sizeof(_api.idx_t) * nvtxs)
 
         with tempfile.TemporaryFile() as tmp:
-            with redirect(sys.stdout, tmp), nogil:
+            with redirect_stdout(tmp), nogil:
                 result = _api.METIS_NodeND(&nvtxs, _xadj, _adjncy, _vwgt, _options,
                                       _perm, _iperm)
```
ysitu: Indentation.

OrkoHunter:
But those tests were failing before this patch. They pass now.

_metis.pyx
```diff
@@ -1,28 +1,32 @@
...
-    """Temporarily redirect operations on file ``source`` to file ``target``.
+def redirect_stdout(target):
+    """Temporarily redirect operations on ``stdout`` to file ``target``.
     """
-    source.flush()
-    fd = source.fileno()
-    with os.fdopen(os.dup(fd), source.mode) as source2:
-        os.dup2(target.fileno(), fd)
+    stdio.fflush(stdio.stdout)
+    with os.fdopen(os.dup(_STDOUT_FILENO), 'w') as source2:
```
ysitu: Rename `source2` to `stdout_backup`.

ysitu:
The tests passing does not mean that any output from METIS is correctly captured. We need to make sure that we relay some meaningful error messages to the user when something goes wrong.

_metis.pyx
OrkoHunter: Done.

_metis.pyx
OrkoHunter: Done.

OrkoHunter:
> We need to make sure that we relay some meaningful error messages to the user when something goes wrong.

That means more test cases.

OrkoHunter:
What are some possible situations when things would go wrong? Partitioning `directed` or `multi` graphs are covered, `nparts > 0` covered, what more could possibly go wrong? We might need another PR for `tests`. As much as I understand, the problem of `StringIO instance of std.stdout` is covered, which is why this PR was created.

ysitu:
I think that you can put some invalid values into `MetisOptions`, e.g., `niter <= 0`. In `options.c` there is a `CheckParams` function. You can find the expected error message there.

OrkoHunter:
Does that mean adding something like `test_MetisOptions` in `test_metis.py`?

```diff
diff --git a/tests/test_metis.py b/tests/test_metis.py
index 337477d..a50a691 100644
--- a/tests/test_metis.py
+++ b/tests/test_metis.py
@@ -2,7 +2,7 @@ import itertools
 import nose.tools
 
 import _metis
-
+import _types
 
 def make_cycle(n):
     xadj = list(range(0, 2 * n + 1, 2))
@@ -68,3 +68,10 @@ class TestMetis(object):
             sorted(map(sorted,
                        [[(sep[0] + i) % n for i in range(1, n // 2)],
                         [(sep[1] + i) % n for i in range(1, n // 2)]])))
+
+    def test_MetisOptions(self):
+        n = 16
+        xadj, adjncy = make_cycle(n)
+        objval, part = _metis.part_graph(
+                xadj, adjncy, 2, options=_types.MetisOptions)
+            
```
Followed by assertions of errors listed in `CheckParams` of `options.c`...

I don't have much idea about using `options` in `part_graph` and other functions.

ysitu:
You create a `MetisOptions` object and set its `niter` property to a negative value.

OrkoHunter:
Simply this?
```py
    def test_MetisOptions(self):
        options = _types.MetisOptions
        options.niter = -1
```
How should I invoke MetisOptions in the functions like `part_graph`?

OrkoHunter:
Well, we should be doing this over #3.

ysitu:
You should do

```python
n = 16
xadj, adjncy = make_cycle(n)
options = types.MetisOptions()
options.niter = -1
assert_raises_regexp(MetisError, '<error message regex>',
                     _metis.part_graph, xadj, adjncy, 2, options=options)
```

OrkoHunter:
For the following,
```py
    def test_MetisOptions(self):
        n = 16
        xadj, adjncy = make_cycle(n)
        options = _types.MetisOptions()
        options.niter = -1
        nose.tools.assert_raises_regexp(_types.MetisError, 'Input Error: Incorrect niter.',
                             _metis.part_graph, xadj, adjncy, 2, options=options)
```
I get
```
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/hunter/GSoC/2015/test/tests/test_metis.py", line 78, in test_MetisOptions
    _metis.part_graph, xadj, adjncy, 2, options=options)
AssertionError: MetisError not raised
```

ysitu:
Was an exception raised at all?

OrkoHunter:
I guess no. It's independent of the asserting string of error.

OrkoHunter:
Error is raised if niter < -1.
This one passes the test and the MetisError is being raised in it.
```py
    def test_MetisOptions(self):
        n = 16
        xadj, adjncy = make_cycle(n)
        options = _types.MetisOptions(niter=-2)
        nose.tools.assert_raises_regexp(_types.MetisError, 'Input Error: Incorrect niter.',
                             _metis.part_graph, xadj, adjncy, 2, options=options)
```

OrkoHunter:
Rebased PR. Updated with

* Fix problem for redirecting of stdout problem
* Renamed `_types` in `_metis.pyx`
* Added the test for checking exceptions while setting invalid values to MetisOptions

networkx/addons/metis/tests/test_metis.py
```diff
@@ -68,3 +68,10 @@ def test_compute_vertex_separator(self):
             sorted(map(sorted,
                        [[(sep[0] + i) % n for i in range(1, n // 2)],
                         [(sep[1] + i) % n for i in range(1, n // 2)]])))
+
+    def test_MetisOptions(self):
+        n = 16
+        xadj, adjncy = make_cycle(n)
+        options = _types.MetisOptions(niter=-2)
+        nose.tools.assert_raises_regexp(_types.MetisError, 'Input Error: Incorrect niter.',
+                             _metis.part_graph, xadj, adjncy, 2, options=options)
```
ysitu: Rename `_types`. Indentation. Does this exceed 79 columns?

OrkoHunter: Yes, it was longer than 79 columns. Updated now.

PR #15, opened 2015-06-25
Generates documentation from docstrings

### LINKED ISSUE #13: Generate documentation from docstrings and make it available on readthedocs.org

OrkoHunter:
Fixes #13 

I've temporarily split the PR into 3 commits. Basically, the last (or last two) commits are to be reviewed.

~~After setting up a hook on readthedocs.org~~, the built docs would be available [here](http://orkohunter-networkx-metis.readthedocs.org/en/doc/) for testing purposes and then finally can be published at 
http://networkx-metis.readthedocs.org/en/latest/

Edit : They are now available at http://orkohunter-networkx-metis.readthedocs.org/en/doc/


OrkoHunter:
The reason why [this page on documentation](http://orkohunter-networkx-metis.readthedocs.org/en/doc/reference/networkx.addons.metis.html) is vacant where it should have been populated by the `automodule` extension, is probably because sphinx is not being able to identify `networkx.addons.metis` as a package. Maybe after `setup.py` installs `networkx-metis`, it would work.

[Build log](https://readthedocs.org/builds/orkohunter-networkx-metis/3005992/)

doc/source/conf.py
```diff
@@ -0,0 +1,298 @@
...
+#needs_sphinx = '1.0'
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+    'sphinx.ext.autosummary',
+    'sphinx.ext.autodoc',
+    'sphinx.ext.intersphinx',
+    'sphinx.ext.viewcode',
```
chebee7i: Do you not need to enable the napolean extension? This will make it so that Sphinx can properly interpret the numpydoc standard.

SanketDG: Yes, and it also renders much nicely with the theme.

OrkoHunter:
After adding `doc/requirements.txt`, the documentation is building in the reference area too. http://orkohunter-networkx-metis.readthedocs.org/en/doc/

doc/source/reference/networkx.addons.metis.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Types
+-----
+
+.. automodule:: networkx.addons.metis.types
+.. autosummary::
+   :toctree: generated/
+
+   MetisOptions
+   MetisError
```
SanketDG: Is this defined in `types.py`?

OrkoHunter: Oh! Didn't add info about `exceptions`... Thanks.

OrkoHunter:
I currently don't have the content for [Who uses NetworkX-Metis](http://orkohunter-networkx-metis.readthedocs.org/en/doc/overview.html#who-uses-networkx-metis) and [Goals](http://orkohunter-networkx-metis.readthedocs.org/en/doc/overview.html#goals) section in the overview. If there's not anything to write in there, we can remove it.

OrkoHunter:
Are we good with this? If there is not urgent need for changes, we can merge this to have our first documentation page published and modify things later after reading it.

chebee7i:
Yeah, let's just get this in. Note, it might be nice to use the IPython directive for documentation.

http://matplotlib.org/sampledoc/ipython_directive.html

OrkoHunter:
That's a good point. Currently we don't have much `examples` in the documentation, but IPython directive should be followed.

SanketDG:
Should the function `MetisRStatus` be documented? Doesn't it acts as a helper function for `MetisError`? Correct me if I'm wrong.

OrkoHunter:
@SanketDG You are right. The `MetisRStatus` class in `enums.py` should be documented but that would be exactly same as what's in the docstring of `MetisError` class in `exceptions.py`

SanketDG:
@OrkoHunter I see. But leaving it blank seems incomplete.

OrkoHunter:
Yes surely. That's bad.

PR #16, opened 2015-06-27
Setup Travis-CI

### LINKED ISSUE #12: Set up Travis CI
Run `nose` and `pylint`; `coverage` is optional.

OrkoHunter:
Fixes #12 .

NOTES

* Currently, instead of `pip install networkx`, I'm installing [this branch](https://github.com/orkohunter/networkx/tree/addons) from my fork of networkx.
* Because of which, `nosetests` is checking the entire `networkx` + `networkx+metis` library.
* Building with branches `master` `setup` and `travis-ci` all merged in one `travis-test` over [https://travis-ci.org/OrkoHunter/networkx-metis](https://travis-ci.org/OrkoHunter/networkx-metis)
* Build status : Passing.

*But it's not ready yet.* Namespace packaging is done, but there seems to be a little problem while installing. `setup.py` is installing a strange file named `_metis.py` in the same location as `_metis.so` with the following content
```py
def __bootstrap__():
   global __bootstrap__, __loader__, __file__
   import sys, pkg_resources, imp
   __file__ = pkg_resources.resource_filename(__name__,'_metis.so')
   __loader__ = None; del __bootstrap__, __loader__
   imp.load_dynamic(__name__,__file__)
__bootstrap__()
```
This file is causing trouble and interfering while importing `networkx.addons.metis._metis`. Renaming the `.so` file doesn't help either, the new file is given the same name except for having `.py` extension.

Related : http://askubuntu.com/questions/442357/environment-setup-for-importing-a-userdefined-module/442363#comment917656_442363

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: python
+
+python:
+  - 2.7
+  - 3.2
+  - 3.3
+  - 3.4
+  - "pypy"
+  - "pypy3"
+
+before_install:
+
+ - sudo apt-get update
+ - sudo apt-get install build-essential
+
+install:
+ - pip install Cython
+ - pip install https://github.com/networkx/networkx/archive/master.zip
+ - pip install .
+
+script:
+ - cd `mktemp -d`  # Navigate into a temporary directory to avoid loading the local code
+ - nosetests networkx.addons.metis.tests.test_metis --nocapture --verbosity=2
+
+notifications:
+  email: false


OrkoHunter:
I'm making a topic in the `cython-users` mailing list. Hope something good comes out of it.

chebee7i:
Is your travis-test branch completely updated? E.g. with master, setup, travis-ci all merged in?  I just tried `python setup.py build` and I complains:

`error: Namespace package problem: networkx.addons.metis is a namespace package, but its
__init__.py does not call declare_namespace()! Please fix it.`

If it is not completely updated, can you push it up so that others can easily debug/test.

OrkoHunter:
@chebee7i Thanks for looking into it. Please give me a moment.

OrkoHunter:
Pushed the changes. It's building for me. Can you please check now?

chebee7i:
Builds fine, but I cannot install via:  `pip install -e .`. That is rather unfortunate. Am I doing anything wrong?

OrkoHunter:
Does your error log contains these lines?
```
/usr/bin/ld: cannot find -lmetis

/usr/bin/ld: cannot find -lgklib

/usr/bin/ld: cannot find -lgklib

/usr/bin/ld: cannot find -lmetis

collect2: error: ld returned 1 exit status

error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

OrkoHunter:
Can you try deleting `networkx/addons/metis/__init__.py` if it's there? And also change the line in `setup.py`
```py
packages = ['networkx.addons.metis', 'networkx.addons', 'networkx'],
```
to
```py
packages = ['networkx.addons', 'networkx'],
```

chebee7i:
The reason `pip install -e .` does not work is because it doesn't install it as a namespace package. If I install using `pip install .`, then it is fine. There is, however, another issue: There are circular imports. `_metis.so` imports `types.py`. But `types.py` needs to import `_metis`. Can you make any of those imports function level imports or create a third file which has the common stuff only?

chebee7i:
Should I still try the change you just suggested? I don't think it is needed.

OrkoHunter:
The circular imports causes problems in the case of
```
from _metis import *
```
Doing absolute module import and using it as `module.name` resolves the problem of circular import. We discussed it here.

chebee7i:
I'm not doing absolute imports, I'm running the code in the branch.

chebee7i:
The circular import issue is still legitimate. Both `_metis.pyx` and `types.py` require that the other be available when they are imported. This will cause a circular import every time.

chebee7i:
The simplest solution is to move

```
class MetisError(Exception):
    """An error type generated from METIS"""
    def __init__(self, rstatus):
        super(MetisError, self).__init__(rstatus)
```

into its own file `exceptions.py`.  Then `_metis.pyx` only needs to import that.  `types.py` will still import `_metis.pyx`, but the circle is broken.

OrkoHunter:
Did your installed `networkx` core package contains `networkx.addons.__init__.py` with 
```py
__import__('pkg_resources').declare_namespace(__name__)
```
in it? If I understand correctly, `networkx-metis` won't work without modifying the `networkx` core package.

OrkoHunter:
The idea of `exceptions.py` sounds reasonably if the problem of circular import is still there.

chebee7i:
I used `exceptions.py` and it fixed the import issue.

My `networkx.addons.__init__.py` in the *core* NetworkX does not have the line you mentioned and it works just fine.

chebee7i:
I sent you a PR...not sure if the travis-test branch is the correct place for that. Feel free to gut the changes and place elsewhere. But I am able to build without issues after that change. I am also using your "addons" branch for NetworkX (which has a blank `networkx.addons.__init__.py`), even though I think you are correct that it is supposed to have that line you mentioned.

OrkoHunter:
It's so astonishing for me that the actual problem (at *this* stage) was still the circular import which I thought was resolved. Thanks @chebee7i.

chebee7i:
Glad to hear. FYI, using absolute imports cannot, in principle, resolve circular imports since a circular import has to do with access at parse time.  Whether you use a relative or absolute import only changes how you access it, but it does not change whether you need access to it in the first place.

ysitu:
With #18, can `networkx/addons/__init__.py` be changed back into an empty file then?

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+        - TEST_PY3="false"
+    - python: 3.2
+      env:
+        - TEST_PY3="true"
+    - python: 3.3
+      env:
+        - TEST_PY3="true"
+    - python: 3.4
+      env:
+        - TEST_PY3="true"
```
ysitu: Add pypy and pypy3.

OrkoHunter:
> With #18, can networkx/addons/__init__.py be changed back into an empty file then?

Things started working after #18. `networkx/addons/__init__.py` was not empty any time. I don't think keeping it empty will help.

OrkoHunter:
Now,
```python
In [1]: import networkx as nx

In [2]: nx
Out[2]: <module 'networkx' from '/usr/local/lib/python2.7/dist-packages/networkx-2.0.dev_20150616095129-py2.7.egg/networkx/__init__.pyc'>

In [3]: import networkx.addons

In [4]: networkx.addons
Out[4]: <module 'networkx.addons' from '/usr/local/lib/python2.7/dist-packages/networkx_metis-1.0-py2.7-linux-x86_64.egg/networkx/addons/__init__.pyc'>
```

After I emptied the `networkx/addons/__init__.py` file

```python
In [1]: import networkx as nx

In [2]: nx
Out[2]: <module 'networkx' from '/usr/local/lib/python2.7/dist-packages/networkx-2.0.dev_20150616095129-py2.7.egg/networkx/__init__.pyc'>

In [3]: import networkx.addons

In [4]: networkx.addons
Out[4]: <module 'networkx.addons' from '/usr/local/lib/python2.7/dist-packages/networkx-2.0.dev_20150616095129-py2.7.egg/networkx/addons/__init__.py'>

In [5]: import networkx.addons.metis
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-5-44df29cab2f4> in <module>()
----> 1 import networkx.addons.metis

ImportError: No module named metis
```

We can see in `Out[4]`, the indentity of `networkx.addons` depends upon the content of `networkx/addons/__init__.py`.

ysitu:
Fair enough. Let's get #18 done to unblock this.

.travis.yml
```diff
@@ -0,0 +1,46 @@
...
+    fi
+
+  - if [[ "${TEST_PY3}" == "true" ]]; then
+      pip install Cython;
+    fi
+
+script:
+
+ - python setup.py build #build networkx-metis
+ - python setup.py test; #test networkx-metis
```
ysitu: Run `python setup.py install` and test the installed copy.

OrkoHunter:
* For python2
 * [Tests fail abruptly](https://travis-ci.org/OrkoHunter/networkx-metis/jobs/68863018#L837).
 * pylint throws negative score of -10.18/10. `_/\_`
* For python3
 * All 5 Tests pass
 * pylint still not pleased. -9.37/10
* For pypy
 * [Tests do not even start](https://travis-ci.org/OrkoHunter/networkx-metis/jobs/68863023#L554)
 * pylint -9.37/10

Latest Build
https://travis-ci.org/OrkoHunter/networkx-metis/builds/68863016

ysitu:
You should be able to get rid of many `no-member` errors by whitelisting `networkx.addons.metis._metis` for importing by pylint.

ysitu:
If setting up pylint is much too tricky, how about trying out landscape.io?

OrkoHunter:
Let me try it.

OrkoHunter:
I would say we go with landscape.io. Can you set up the hook there?

ysitu:
I am on an airplane right now.

OrkoHunter:
Yeah you must be going.

chebee7i:
Anything that requires admin level privileges will require action from @hagberg, @dschult, or @jtorrents .

hagberg:
Let's add @chebee7i as owner - I can't do it right now @dschult, @jtorrents?

jtorrents:
Done! @chebee7i is in the owner group now. I'm not sure how to set up the travis hook (still have to set the readthedocs for my fork). So I'll let @chebee7i do it.

chebee7i:
Done for both metis and lemon. @OrkoHunter I need your readthedocs username as well.

OrkoHunter:
My username for readthedocs is same as GitHub `orkohunter`. However, my fork of docs is here
http://orkohunter-networkx-metis.readthedocs.org/en/doc/

chebee7i:
It is case-sensitive...that is why I couldn't add you. :)

OrkoHunter:
Oh.. My bad then. :)

OrkoHunter:
Triggered first build with python 2.7, 3.2+, pypy and pypy3. Let's wait for the results.

OrkoHunter:
Here it is, ironically it **fails for Python 2.7** and *passes for Python 3.2+, pypy and pypy3*. The problem lies with *installation*.
Right now, it is
```sh
python setup.py install
python setup.py test
```
But if we skip the installation and rather only build and test
```sh
python setup.py build
python setup.py test
```
it passes for all.

OrkoHunter:
Well it passed for `build` and `test` as I mentioned earlier.

ysitu:
With `python setup.py test` you are not testing the installed code.

OrkoHunter:
To test the installed code, do we need to run a python script which would do some imports and runs some tests within it?

ysitu:
`cd` into some unrelated directory and run `nosetests`?

OrkoHunter:
`nosetests` searches recursively for all the tests within a directory. Won't making a new unrelated directory will run 0 tests? And running `nosetests` from the top level will also be same as `python setup.py test`, as both are generating the same OS error. I'm still surprised the variation of running the test between Python 2 and 3.

ysitu:
You can supply a module name to `nosetests`?

OrkoHunter:
Yes, like `nosetests only_this_test.py`.

chebee7i:
Look at what we do in networkx.

.travis.yml
```diff
@@ -0,0 +1,41 @@
...
+
+install:
+  - pip install Cython
+  - pip install https://github.com/networkx/networkx/archive/master.zip
+  - python setup.py install
+
+script:
+ - mkdir new
+ - cd new
+ - nosetests ../networkx/addons/metis
```
ysitu: What does using the `new` directory achieve if you continue to ask `nosetests` to use the checked out repository code?

.travis.yml
```diff
@@ -0,0 +1,41 @@
...
+
+
+before_install:
+
+ - sudo apt-get update
+ - sudo apt-get install build-essential
+
+install:
+  - pip install Cython
+  - pip install https://github.com/networkx/networkx/archive/master.zip
```
ysitu: You should not need this now that you have `networkx` in `install_requires`.

.travis.yml
OrkoHunter: The written tests inside `networkx/addons/metis/tests/test_metis.py` imports `_metis` and `types` as
```
from networkx.addons.metis import _metis
```
This is an absolute import. In my opinion, the module is being imported from the installed packages directory instead of the local code.

.travis.yml
OrkoHunter: But that searches `networkx` on pypi. The current master is not uploaded on pypi, so don't we need the github version of networkx to install?

.travis.yml
ysitu: But your repository is structured exactly the way that makes such imports possible.

.travis.yml
ysitu: If you need to test again the `networkx` master, then the user will have to use `networkx-metis` with the `networkx` master. This is just ugly. We need to enable 1.10 now.

OrkoHunter: 1.10rc2 does not contain the `networkx/addons` which an add-on needs. We'll have to wait for a new release to depend upon `install_requires`. Till then, this should be the way, and it's just ugly.

.travis.yml
OrkoHunter: All I can currently think of is inside `new`, we run our script to check the installed code. If nothing else is suggested, should I go with this?

ysitu: The command to run is `nosetests networkx.addons.metis.tests`. Ideally you should create a temporary directory elsewhere instead of under the checkout directory.

.travis.yml
ysitu: We should make the case that the final 1.10 release should contain the necessary machinery.

.travis.yml
chebee7i: You can always print out the imported module to see if it is, in fact, being imported from the installed location or not.

OrkoHunter: @chebee7i Nice checking idea. Let me try that.

.travis.yml
OrkoHunter: Yes, that seems to be appropriate. @chebee7i ?

.travis.yml
OrkoHunter: The tests are importing modules from the installed package in `site-packages`.
https://travis-ci.org/networkx/networkx-metis/jobs/69095104#L798
```sh
$ ipython -c 'from networkx.addons import metis; metis'
Out[1]: <module 'networkx.addons.metis' from '/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/networkx_metis-1.0-py3.4-linux-x86_64.egg/networkx/addons/metis/__init__.py'>
```
No worries then.

OrkoHunter:
pypy needs special installation http://pypy.readthedocs.org/en/latest/faq.html#module-xyz-does-not-work-with-pypy-importerror

OrkoHunter:
Green tick feels :)
* All the 5 Tests are passing for
 * Python 2.7
 * Python 3.2, 3.3 and 3.4
 * pypy and pypy2

* We can be sure that the installed code is being tested and not just local code. https://github.com/networkx/networkx-metis/pull/16#discussion_r33654384

* We are currently installing networkx bundle from the github repository because pip version of networkx does not include this patch https://github.com/networkx/networkx/pull/1595. So we have two options
 1. Wait for the next release of networkx with https://github.com/networkx/networkx/pull/1595 in it and remove [this line](https://github.com/networkx/networkx-metis/pull/16/files#diff-354f30a63fb0907d4ad57269548329e3R18) at that time because `install_requires` will automatically do the job.

 2. Publish https://github.com/networkx/networkx/pull/1595 in the latest release of networkx so that networkx-metis won't have to wait for long to be used by real users. (We are pretty much good to go)

chebee7i:
Note, to guarantee that the (all) tests are using the installed networkx rather than the repository one, you can move directories...but also, each test file must have `from __future__ import absolute_import`.

OrkoHunter:
Should `from __future__ import absolute_import` be added in the test files with this PR? Tests are to be re-written (#26) so maybe then.
>  you can move directories...

Didn't get what this line is really asking for. Or is it?

chebee7i:
I'm just mentioning that if the package directory is not in the current directory, then you can guarantee that a local import does not happen.

chebee7i:
As to whether the `absolute_import` line should be added here or elsewhere, it's up to you...in the end, the effect is the same.

OrkoHunter:
Oh yes well. We are assured about our current situation that it imports installed networkx (from `site-packages`), aren't we? https://github.com/networkx/networkx-metis/pull/16#discussion_r33654384

OrkoHunter:
Regarding landcape.io, I don't know whatever the *bug* they are facing as even the first check has not started yet. https://landscape.io/github/networkx/networkx-metis
I've filed an issue and sent a mail regarding this. I don't think there are is any configuration settings that you'd have missed while adding `networkx-metis` there. So, most probably it's their problem.

chebee7i:
What bug? It says 97% passed.

OrkoHunter:
Wow! This is amazing. Half an hour ago the page was total blank.

.travis.yml
ysitu: You can move the check into the setup code of the test class.

ysitu: I mean printing out the module for manual inspection.

networkx/addons/metis/tests/test_metis.py
```diff
@@ -18,6 +17,12 @@ def make_cycle(n):
 class TestMetis(object):
 
     def test_node_nd(self):
+        #Check for the import location
+        print(exceptions.__file__)
+        #Check for the import location
+        print(_metis.__file__)
+        #Check for the import location
+        print(types.__file__)
```
OrkoHunter: @ysitu These hackery gave this output on the travis
```
$ nosetests --nocapture ../networkx/addons/metis/tests/
/home/travis/build/OrkoHunter/networkx-metis/networkx/addons/metis/exceptions.py
/home/travis/build/OrkoHunter/networkx-metis/networkx/addons/metis/_metis.so
/home/travis/build/OrkoHunter/networkx-metis/networkx/addons/metis/types.py
```
I don't think it's testing from the installed location. On the other hand, I'm also not much worrying about it, because the installed code is in no way going to be different from what is inside `build` because the same is copied while installation. What do you say? I say this is good enough, probably not the best.

ysitu:
Your output is exactly what I expected.

Installation is the very first part of the user experience of this package. We need to get testing of that right. This package is more structurally complex than many others. It has an extension built from Cython, and the built shared object needs to live in a different directory than the source file (although I think that `_metis.pyx` and `_api.pxd` should be moved into `networkx/addons/metis`). We need rigorous testing to ensure that no surprises occur after installation.

OrkoHunter:
I tried moving the tests on the top level (out of networkx directory) thinking that it won't recognize the local networkx but it is doing the same. In fact nosetests tend to find the packages locally first that's why it's getting hard to check installation from nosetests.
I think we need some other script than `test_metis.py` to check installation only. And yes, I too feel now that installation must be tested rigorously.

chebee7i:
I'd like to see `_metis.pyx` and `_api.pxd` moved as well.

This is potentially relevant for the nosetests: https://github.com/pypa/pip/issues/3#issuecomment-4002127

@OrkoHunter when you install networkx-metis, where does it go? Can you not move to that directory where `metis` is a subdirectory and then run nosetests. It seems like it will be alone and there is no risk that tests from other packages will be run. The only issue is whether nose will still discover the tests (see the link I included).

ysitu:
Use `mktemp -d` to create a temporary directory and `cd` into it before testing. Pass a dot-separated module name instead of a path to `nosetests`.

OrkoHunter:
> @OrkoHunter when you install networkx-metis, where does it go? Can you move to that directory where metis is a subdirectory and then run nosetests. It seems like it will be alone and there is no risk that tests from other packages will be run.

It installs into the `site-packages` directory of the respective `python-x.y` directory. I'll have to make cases for each version of python.

@ysitu I'll try that.

OrkoHunter:
This is something I tried.
```yaml
 - dir=`mktemp -d` && cd $dir
 - wget https://raw.githubusercontent.com/OrkoHunter/networkx-metis/travis-ci/networkx/addons/metis/tests/test_installation.sh
 - sudo chmod +x test_installation.sh
 - ./test_installation.sh
```
The script contains this
```sh
set -e #It will immediately stop your script if a simple command fails.

{ #this is similar to a `try` block

    python -c '
try:
    from networkx.addons import metis
    print(metis.__file__)
except:
    raise SystemExit("ImportError! NetworkX-Metis not installed properly.")
' && echo "Import successful!"

} || { #this is similar to a `catch` block

    echo "ImportError: Module not found"

}
```
And the output on travis is
```
$ ./test_installation.sh
/home/travis/build/networkx/networkx-metis/networkx/addons/metis/__init__.py
Import successful!
```
I don't simply understand why is the script present inside `/tmp/tmp.xxxxx/test_installation.sh` is importing modules from. from `/home/travis/build`..

OrkoHunter:
Well, I manually removed `/home/travis/build` and similar paths from `sys.path`. Lets see how it imports the local code now.

OrkoHunter:
> Pass a dot-separated module name instead of a path to nosetests.

```sh
$ nosetests networkx.addons.metis

Ran 0 tests in 0.000s

OK
$ nosetests networkx/addons/metis             [master] 10:35:30
.....
Ran 5 tests in 0.029s

OK
networkx-metis/                                               [master] 10:35:35
```
I'm not sure about the passing arguments to nosetests. It's rather [optional](http://nose.readthedocs.org/en/latest/usage.html#basic-usage), nosetests itself finds out the test modules. So, I go with one which works.

OrkoHunter:
> Lets see how it imports the local code now.

Well it doesn't import the local code now. Nor does it import the system code even if `site-packages` is present in sys.path.
Captured from log
```sh
/home/travis/virtualenv/python2.7.9/lib/python27.zip
/home/travis/virtualenv/python2.7.9/lib/python2.7
/home/travis/virtualenv/python2.7.9/lib/python2.7/plat-linux2
/home/travis/virtualenv/python2.7.9/lib/python2.7/lib-tk
/home/travis/virtualenv/python2.7.9/lib/python2.7/lib-old
/home/travis/virtualenv/python2.7.9/lib/python2.7/lib-dynload
/opt/python/2.7.9/lib/python2.7
/opt/python/2.7.9/lib/python2.7/plat-linux2
/opt/python/2.7.9/lib/python2.7/lib-tk
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages
ImportError! NetworkX-Metis not installed properly.
ImportError: Module not found
```

ysitu:
What if you pass `networkx.addons.metis.test_metis` to `nosetests`?

OrkoHunter:
`tests` is not a module. and `test_metis` won't be recognized as a module (maybe because the absence of `__init__.py` within the directory)

However, `networkx.addons.metis.tests`, `networkx.addons.metis.test_metis` and `networkx.addons.metis.tests.test_metis` all end up with similar results.
```sh
$ nosetests networkx.addons.metis.test_metis
E
======================================================================
ERROR: Failure: AttributeError ('module' object has no attribute 'test_metis')
  File "/usr/lib/python2.7/dist-packages/nose/loader.py", line 400, in loadTestsFromName
    module = resolve_name(addr.module)
  File "/usr/lib/python2.7/dist-packages/nose/util.py", line 321, in resolve_name
    obj = getattr(obj, part)
AttributeError: 'module' object has no attribute 'test_metis'
-------------------- >> begin captured logging << --------------------
root: INFO: Generating grammar tables from /usr/lib/python2.7/lib2to3/Grammar.txt
root: INFO: Generating grammar tables from /usr/lib/python2.7/lib2to3/PatternGrammar.txt
--------------------- >> end captured logging << ---------------------

Ran 1 test in 0.001s

FAILED (errors=1)
```

ysitu:
Even after you add an `__init__.py` to make `tests` a module?

OrkoHunter:
No.
```sh
$ touch networkx/addons/metis/tests/__init__.py
$ nosetests networkx.addons.metis.tests.test_metis 
.....
Ran 5 tests in 0.003s

OK
```

chebee7i:
> It installs into the site-packages directory of the respective python-x.y directory. I'll have to make cases > for each version of python. But that would be .egg file. I wonder how can I cd into it.

Yes, this is always true. My question was how the namespace handled. In any event, look at the `.travis.yml` for `networkx`. This problem (of finding the path) has been solved already. There are other things it does that you will also need to do.

ysitu:
If the `tests` directory not having `__init__.py` gets in the way of testing, we should add that file.

OrkoHunter:
##### A short report after some long hours of debugging and trying things out (a great time sink)

`NetworkX` uses a very nice way of testing the installed code. With these two lines
```
export NX_INSTALL=`pip show networkx | grep Location | awk '{print $2"/networkx"}'`
cd $NX_INSTALL
```
it migrates the path to `.egg` directory inside the `site-packages` and runs `nosetests` there.

*So, why don't we simply use it for `networkx-metis`?*

Taken from http://svn.python.org/projects/sandbox/trunk/setuptools/doc/formats.txt

> There are two basic formats currently implemented for Python eggs:
>
> 1. ``.egg`` format: a directory or zipfile *containing* the project's
   code and resources, along with an ``EGG-INFO`` subdirectory that
   contains the project's metadata
> 
> 2. ``.egg-info`` format: a file or directory placed *adjacent* to the
   project's code and resources, that directly contains the project's
   metadata.

So, I tried three different builds [#113](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69556504) [#114](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69557106) and [#115](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69558906) with `pip install -e .`, `pip install .` and `python setup.py install` respectively. and got three different results in all three cases.

1. In first case, an [`.egg-link`](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69556504#L766) is generated and is installed inside `site-packages`. It's just a symbolic link to where the code is actually building, `/home/travis/build` in that case. So, there is no installed code inside `site-packages` for networkx-metis to test.

2. In second case, a [`.egg-info`](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69557106#L3347) and a [`-nspkg.pth`](https://travis-ci.org/OrkoHunter/networkx-metis/builds/69557106#L3348) files are being created. `.pth` file simply adds the building directory's path inside `sys.path`. So, it also acts as a symlink for the actual `build` directory.

3. This one creates a [`.egg`](https://travis-ci.org/OrkoHunter/networkx-metis#L1092) file inside site-packages. But unlike the case of `NetworkX` which has a directory type `.egg` inside `site-packages`, `Networkx-Metis` seems to be generating an archive `.egg`. The single line of code which I mentioned above won't work for this.

So, the possible way I can think of is to write script to extract the archieve, `cd` into it and then test it.

OrkoHunter:
I have updated the script. It needs #33 in to run.

chebee7i:
So on my machine, after installing via `pip install .`, the metis stuff is literally put into `site-packages/networkx/addons/metis`.  Is that not the same for you?

OrkoHunter:
Well yeah. It's there. Although the first two `__init__.py` files (`networkx/`, `networkx/addons`) are missing.
Didn't see it. I'll use `pip install .` then.

chebee7i:
This file should exist right? `/path/to/site-packages/networkx/__init__.py` It should exist b/c `networkx` is a valid Python package.

However, `/path/to/site-packages/networkx/addons/__init__.py` probably does not exist, as this would need to be provided by NetworkX 1.10. Is that correct?

OrkoHunter:
Can you try doing an import?
```
from networkx.addons import metis
```

chebee7i:
Uhoh, this is not good. I made a clean virtualenv.  Then:

```bash
pip install Cython
git clone https://github.com/networkx/networkx
cd networkx
git checkout v1.10
pip install .
cd ../
git clone https://github.com/networkx/networkx-metis
pip install .
```

This breaks Python altogether for me.

```
  File "/home/me/.virtualenvs/demo/lib/python2.7/site.py", line 703, in <module>
    main()
  File "/home/me/.virtualenvs/demo/lib/python2.7/site.py", line 683, in main
    paths_in_sys = addsitepackages(paths_in_sys)
  File "/home/me/.virtualenvs/demo/lib/python2.7/site.py", line 282, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "/home/me/.virtualenvs/demo/lib/python2.7/site.py", line 204, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "/home/me/.virtualenvs/demo/lib/python2.7/site.py", line 173, in addpackage
    exec(line)
  File "<string>", line 1, in <module>
KeyError: 'networkx'
```

OrkoHunter:
Why did you checkout `v1.10`? We need to install the master of `NetworkX` I think.

chebee7i:
I'll do that now. I just wanted to see what would happen if kept 1.10 as is (without addon support) and then installed networkx-metis. It breaks in a very bad way. Users cannot even uninstall networkx-metis.

OrkoHunter:
Oh. Got your point.

chebee7i:
From master, it works just fine.

chebee7i:
So I think this really points out that either we should update 1.10 to include the addons or we do not release networkx-metis until 2.0.

OrkoHunter:
Yes, exactly!

hagberg:
Remind me why we need this namespace trickery anyway?  If forgotten what the advantages are here.

OrkoHunter:
That is because `networkx` and `networkx-metis` are sharing a common namespace `networkx.addons`. So, while doing
```
from networkx.addons import metis
```
If python searches inside the `networkx-2.0-dev-xyz.egg` first and didn't find any `networkx.addons` with decalred namesapce from `pkg_resources`
```
__import__('pkg_resources').declare_namespace(__name__)
```
It will immediately stop the search and will never be able to find `networkx-metis`.

chebee7i:
So once networkx has `networkx/addons/__init__.py` and that file declares it to be a namespace, then anyone can create namespace packages that will be importable as `from networkx.addons import myextension`.  No further changes are required from `networkx`.

If we did not use namespaces, then we'd have to modify networkx every time a new extension were created by adding stubs such as:

```
# networkx/addons/metis.py
from nonnamespaced_metis import *
```

Then users might use these as `from networkx.addons import metis`.

OrkoHunter:
Yes. Having the `networkx/addons/__init__.py` is basically the *Implementing add-on system for NetworkX*.

hagberg:
No, I understand how it works.  I am asking *why* we want to do it this way as opposed to a separate package.  Maybe this was discussed before and I have forgotten the reasons.

chebee7i:
Both options seem fine to me. The namespace route makes it more flexible in the sense that people developing addons don't have to get official stubs added. This would mean that new extensions could only be released when networkx has a new release...which seems undesirable. If an extension is no longer available, then we'd have a stub file that no longer makes sense until a new release of networkx.

chebee7i:
As far as I know, this flexibility is the *only* benefit over making each extension their own, independent package.  All of this assumes we do want stuff available within `networkx.addons`. If we don't want that, then independent packages is the only option that makes sense.

hagberg:
This is probably the wrong place for this discussion but I'm just trying to remember why we need these packages in the networkx namespace.  It seems like potential difficulty and confusion in installation and testing. so I'd like to remember what we gain.  If I remember right SciPy rejected this approach (for scikits) and I have abandoned it before too in favor of separate packages.

chebee7i:
Agreed on the potential downsides. While namespaces are officially supported, they do just seem messy. Are there any examples of big packages in Python using namespaces successfully and happily?

OrkoHunter:
If we consider the upcoming addon `networkx-lemon`, I think we do *gain* having it under `networkx.addons`. 

For users who have a good habit of 
```
networkx.some_slow_func(over_large_graph)
```
They can make a small change in their library
```py
from networkx.addons import lemon
lemon.same_func_incredibly_fast(over_same_graph)
```
and be benefited. The `lemon.some_func` is designed in the same way as `nx.some_func` because we don't want the two to be potentially confusing.

The `namespace` way seems to be more appealing because we'll be able to use stuff from `networkx.addons`. Having an import of `networkx-lemon` and using some of it's function as well as interfacing it with a function of `networkx` will not be much appreciated if they do not come from a same namespace.

hagberg:
But couldn't you do the following (or similar) with a separate package - no more complex
```py
import networkx-lemon
networkx-lemon.same_func_incredibly_fast(over_same_graph)
```

OrkoHunter:
Agreed, it's similar. I just think a user will hesitate taking an output generated from `networkx-lemon` and using it further with `networkx`. These two imports highlights both as separate package, even if the first one is written *solely* written for the latter.

OrkoHunter:
If I am correct about sckit-learn, it started in 2007, while PEP 420 (which we are using) was proposed in 2012. I do not think SciPy had PEP 420 then, to reject.

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+
+install:
+ - pip install Cython
+ - pip install https://github.com/networkx/networkx/archive/master.zip
+ - python setup.py build
+ - pip install .
+
+script:
+ - export NX_INSTALL=`pip show networkx | grep Location | awk '{print $2"/networkx"}'`
+ - cd $NX_INSTALL  # Goes inside the directory where NetworkX is installed
```
ysitu: Does it work if you go into a random directory (with #33 merged)?

OrkoHunter: The main motive of this is to be able to reach to `site-packages` of the particular version's environment. How can I test the installed code from a random directory?

OrkoHunter: Btw., all tests are passing now and the imports are from the installed location.

ysitu: You have only one version installed. If the tests run, they run from the correct location. Going into a random directory merely serves to hide the checked out code from sight.

OrkoHunter: This is a relevant part of the log.
```
$ nosetests networkx.addons.metis.tests.test_metis --nocapture --verbosity=2
networkx.addons.metis.tests.test_metis.TestMetis.test_MetisOptions ... ok
networkx.addons.metis.tests.test_metis.TestMetis.test_compute_vertex_separator ... ok
networkx.addons.metis.tests.test_metis.TestMetis.test_node_nd ... 
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/networkx/addons/metis/exceptions.pyc
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/networkx/addons/metis/_metis.so
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/networkx/addons/metis/types.pyc
ok
networkx.addons.metis.tests.test_metis.TestMetis.test_part_graph ... ok
networkx.addons.metis.tests.test_metis.TestMetis.test_selfloops ... ok
```
I'll remove the lines I added for testing purpose when you say this is ready to be merged.

ysitu:
I think that the detours in this Issue have exaggerated the complexity of namespace packages. By now we have determined, AFAICT, that the technical solution is not that complicated.

Yingchong Situ

.travis.yml
OrkoHunter: Got it now. Yes that works from a temp directory.

OrkoHunter:
AFAIK, the amount of dependency of networkx upon its addons in the proposals have considerably been decreased through out time.

The first implementation proposed by @ysitu was to manage the external code and wrappers in `networkx.external` and have python code inside `networkx.algorithms`.

Later on, it was decided that the networkx repository shouldn't take source code of external libraries. So, the idea of hosting separate `networkx-<addons>` under networkx umbrella came up. And having a `<addon>.py` inside `networkx/addon/` to do a wildcard import from the repository was considered. But this was still a bit useless noise inside networkx repository.

So, we moved to use namespace packages with the simplest version of addon system by declaring `networkx.addons` a shared namespace.

OrkoHunter:
The previous build passed successfully while importing stuff from installed location. I have removed the lines written for testing purposes. This PR now activates Travis-CI for this repository.

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+
+before_install:
+
+ - sudo apt-get update
+ - sudo apt-get install build-essential
+
+install:
+ - pip install Cython
+ - pip install https://github.com/networkx/networkx/archive/master.zip
+ - python setup.py build
```
ysitu: Do you need this? If yes, this will be problematic for installing from PyPI.

OrkoHunter: It didn't work on my local computer. I've removed the line to see what happens.

OrkoHunter: Tests pass without explicitly `build`. pip doesn't support a build command, so it's include inside install.

PR #17, opened 2015-06-27
Identation fix in types.py

OrkoHunter:
Wow, theses little things I didn't notice. Thanks Travis.

### DIFF
--- a/networkx/addons/metis/types.py
+++ b/networkx/addons/metis/types.py
@@ -47,7 +47,7 @@ class MetisOptions(object):
             vol Total communication volume minimization
             === =======================================
         """
-    return self._objtype
+        return self._objtype
 
     @objtype.setter
     def objtype(self, value):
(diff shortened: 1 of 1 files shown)


chebee7i:
Is this going into the right repository? Looks like it is going into networkx:master.

OrkoHunter:
Well this PR is #17 of `networkx-metis` seeing from URL. And `networkx` doesn't even have any `networkx.addons.metis.types`. I think the scheme of that is <username>:<branch> and <organization>:<branch>. Nothing to do with repository in both case.

chebee7i:
Strange that I never noticed it before...and it indeed the same on other PRs.   But it still seems out.  The syntax is:   `repositoryname:branchname`. So when it says:

	OrkoHunter  wants to merge 1 commit into networkx:master from OrkoHunter:types_fix

the definitely seems "off".

chebee7i:
In the repository settings, is there some way to specify this and it is just specified wrong?

chebee7i:
Looks like it is ubiquitous.  The format is actually `username:branchname`. So like you said, the repository name is not part of it.

OrkoHunter:
Maybe it's a new change? That's why we never noticed it earlier.

PR #18, opened 2015-06-28
Fix circular import

OrkoHunter:
This includes the changes sent in to `travis-test` branch by @chebee7i .

@ysitu I think I'm clear now about the presence of `__init__.py` in `networkx/addons/metis/`. It is necessary. See, networkx and networkx-metis are sharing the common namespace `networkx.addons`. So, in both the directories there should be an `__init__.py` containing the line 
```py
__import__('pkg_resources').declare_namespace(__name__)
```
and *nothing* else. This establishes their relation of sharing a namespace. Any other package in future also needs the same configuration.

Now, if `networkx` and `networkx-metis` are both installed in the system, `networkx.addons` is a package but `networkx.addons.metis` is not because of the absence of `__init__.py`. So, that's why we need it.

I'll create a `test` branch shortly with current `master` `setup` and this `fix_circular_import`. We can try installing that branch.

### DIFF
--- a/networkx/addons/metis/exceptions.py
+++ b/networkx/addons/metis/exceptions.py
@@ -1,16 +1,20 @@
+from networkx.addons.metis import types
 __all__ = ['MetisError']
+
 
 class MetisError(Exception):
     """Error returned by METIS"""
 
-    def __init__(self, rstatus):
-        """Types of return codes.
+    def __init__(self, rstatus, msg):
+        """Initializes a MetisError object.
 
-            ======================================== ==================
-            Returned normally                        METIS_OK
-            Returned due to erroneous inputs/options METIS_ERROR_INPUT
-            Returned due to insufficient memory      METIS_ERROR_MEMORY
-            Some other errors                        METIS_ERROR
-            ======================================== ==================
+        Parameters
+        ----------
+        rstatus : int
+            Error code returned by METIS.
+
+        msg : string
+            Error message returned by METIS.
         """
-        super(MetisError, self).__init__(rstatus)
+        super(MetisError, self).__init__('{0}: {1}'.format(
+            types.MetisRStatus(rstatus).name, msg))
--- a/networkx/addons/metis/exceptions.py
+++ b/networkx/addons/metis/exceptions.py
@@ -1,0 +1,16 @@
+__all__ = ['MetisError']
+
+class MetisError(Exception):
+    """Error returned by METIS"""
+
+    def __init__(self, rstatus):
+        """Types of return codes.
+
+            ======================================== ==================
+            Returned normally                        METIS_OK
+            Returned due to erroneous inputs/options METIS_ERROR_INPUT
+            Returned due to insufficient memory      METIS_ERROR_MEMORY
+            Some other errors                        METIS_ERROR
+            ======================================== ==================
+        """
+        super(MetisError, self).__init__(rstatus)
--- a/networkx/addons/metis/metis.py
+++ b/networkx/addons/metis/metis.py
@@ -14,6 +14,7 @@ import sys
 
 import networkx as nx
 
+from networkx.addons.metis import exceptions
 from networkx.addons.metis import _metis
 from networkx.addons.metis import types
 __all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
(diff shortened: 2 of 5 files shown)


networkx/addons/metis/exceptions.py
```diff
@@ -0,0 +1,6 @@
+__all__ = ['MetisError']
+
+class MetisError(Exception):
+    """An error type generated from METIS"""
```
ysitu: `Error returned by METIS.` Add a blank line before `def __init__()`.

networkx/addons/metis/exceptions.py
```diff
@@ -0,0 +1,6 @@
...
+    def __init__(self, rstatus):
+        super(MetisError, self).__init__(rstatus)
```
ysitu: Add docstring. Document type and values of `rstatus`.

networkx/addons/metis/metis.py
```diff
@@ -89,7 +90,7 @@ def _convert_exceptions(func, *args, **kwargs):
 
 @nx.utils.not_implemented_for('directed')
 @nx.utils.not_implemented_for('multigraph')
-@_convert_exceptions(nx.NetworkXError, (ValueError, TypeError, types.MetisError))
+@_convert_exceptions(nx.NetworkXError, (ValueError, TypeError, exceptions.MetisError))
```
ysitu: This is wider than 79 columns.

networkx/addons/metis/exceptions.py
OrkoHunter: Added.

networkx/addons/metis/exceptions.py
OrkoHunter: Done.

networkx/addons/metis/exceptions.py
```diff
@@ -0,0 +1,16 @@
+__all__ = ['MetisError']
+
+class MetisError(Exception):
+    """Error returned by METIS"""
+
+    def __init__(self, rstatus):
```
ysitu: Now I notice that `MetisError` is not used this way in `check_result` in `_metis.pyx`. Currently `check_result` is defined like this:

```cython
    if result != _api.METIS_OK:
        raise types.MetisError(msg)
```

`msg` is a string instead of a `MetisRStatus` enum. We certainly need to pass on the error message. The error code seems redundant to me, although if for any reason we fail to capture a meaningful error message, it can serve provide some high-level hint at what happened. What do you think?

OrkoHunter: The error code is really not at par with the error messages generated by METIS. Are you saying that we should also provide the code too along with the message?

> if for any reason we fail to capture a meaningful error message, it can serve provide some high-level hint at what happened

Well, I can't think of the reasons of message capturing failure. So, I'm unable to comment.

ysitu: Yes, relaying both the error code and the error message is what I meant. I think that capturing stdout of METIS is a fragile practice. We have not audited (will we ever?) METIS to ensure that it always writes something meaningful to stdout. If it does not, then the error code alone still adds some value compared to a bare `MetisError` object.

OrkoHunter: Okay now I'm to in favor or relaying the error code along with the message. We'll never come to know what METIS returns in the error cases. Or should we wait for the first error message if anyone ever mess things up and report to us in future? If not, you can suggest me how to do this so that we can get this merged soon.

ysitu: We do what we can now. You can use something along the line of

```python
def MetisError(Exception):

    def __init__(self, rstatus, msg):
        super(MetisError, self).__init__('{0}: {1}'.format(
            types.MetisRStatus(rstatus).name, msg))
```

networkx/addons/metis/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, rstatus, msg):
+        """Types of return codes.
+
+            ======================================== ==================
+            Returned normally                        METIS_OK
+            Returned due to erroneous inputs/options METIS_ERROR_INPUT
+            Returned due to insufficient memory      METIS_ERROR_MEMORY
+            Some other errors                        METIS_ERROR
+            ======================================== ==================
+        """
```
ysitu: Now this docstring needs adapting. I think that you can remove this table and document `rstatus` and `msg` the usual way:

```python
"""Initializes a MetisError object.

Parameters
----------
rstatus : int
    Error code returned by METIS.

msg : string
    Error message returned by METIS.
"""
```

networkx/addons/metis/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+        """Initializes a MetisError object.
+
+        Parameters
+        ----------
+        rstatus : int
+            Error code returned by METIS.
+
+        msg : string
+            Error message returned by METIS.
+        """
```
OrkoHunter: Looks good.

PR #25, opened 2015-06-29
Change import routine

OrkoHunter:
Let's change the way of using the functions from
```python
>>> from networkx.addons.metis import metis
>>> metis.partition
<function networkx.addons.metis.metis.partition>
```
to
```python
>>> from networkx.addons import metis
>>> metis.partition
<function networkx.addons.metis.metis.partition>
```

The later one makes more sense.


OrkoHunter:
This will be similar for lemon.
```py
>>> from networkx.addons import lemon
>>> lemon.some_shortest_path_func()
```

ysitu:
Does it make sense to simply rename `metis.py` to `__init__.py`?

OrkoHunter:
Having a `metis.py` to use all the wrapper functions from `_metis.pyx` makes more sense to me. However, we can do the renaming trick but there would be less of a *wrapper module*.

ysitu:
I am not sure if I understand what you mean by "there would be less of a wrapper module". I just find the extra level of indirection unnecessary if `__init__.py` does only `from .metis import *`.

OrkoHunter:
Ah well, let's change it.

OrkoHunter:
Removed `metis.py`

Issue #26, opened 2015-06-29
Change test_metis.py to test the exposed functions


ysitu:
We should not be directly testing functions from `_metis.pyx` as those are implementation detail.

OrkoHunter:
Okay.. So rather we'd be testing the functions in the top level of `networkx.addons.metis`.

ysitu:
That is about the philosophy of testing. Impose constraints on only the public interfaces so that you are handcuffed to the implementation details.

ysitu:
@OrkoHunter Can you prioritize this?

OrkoHunter:
Sure. This is a must. I'll be sending a PR shortly.

PR #27, opened 2015-06-30
Some style fixes

OrkoHunter:
Fix https://landscape.io/github/OrkoHunter/networkx-metis/3/messages/error
* binary-op-exception (W0711):
 	Exception to catch is the result of a binary “%s” operation Used when the exception to catch is of the form “except A or B:”. If intending to catch multiple, rewrite as “except (A, B):”

It seems that populating `networkx/addons/metis/__init__.py` has generated a couple of cyclic imports.
https://landscape.io/github/OrkoHunter/networkx-metis/3/messages/smell

~~They can be fixed by using `from . import module` which I think is okay to do.~~
Nope. Can't fix that.

### DIFF
--- a/networkx/addons/metis/__init__.py
+++ b/networkx/addons/metis/__init__.py
@@ -81,7 +81,7 @@ def _convert_exceptions(convert_type, catch_types=None):
     def _convert_exceptions(func, *args, **kwargs):
         try:
             return func(*args, **kwargs)
-        except catch_types or () as e:
+        except catch_types as e:
             exc = e
         except Exception as e:
             if catch_types is not None:
@@ -255,8 +255,9 @@ def partition(G, nparts, node_weight='weight', node_size='size',
         tpwgts = list(itertools.chain.from_iterable(tpwgts))
 
     with _zero_numbering(options):
-        objval, part = _metis.part_graph(xadj, adjncy, nparts, vwgt, vsize, adjwgt,
-                                         tpwgts, ubvec, options, recursive)
+        objval, part = _metis.part_graph(xadj, adjncy, nparts, vwgt, vsize,
+                                         adjwgt, tpwgts, ubvec, options,
+                                         recursive)
 
     parts = [[] for i in range(nparts)]
     for u, i in zip(G, part):


networkx/addons/metis/__init__.py
```diff
@@ -81,7 +81,7 @@ def function():
     def _convert_exceptions(func, *args, **kwargs):
         try:
             return func(*args, **kwargs)
-        except catch_types or () as e:
+        except(catch_types, ()) as e:
```
chebee7i: I think you still want a space after except. This isn't a function call. http://stackoverflow.com/a/6470452

OrkoHunter: Well yes. Updating.

SanketDG:
Line 258 of this file and line 55 of test_metis.py longer than 79 chars. Might want to fix that.

OrkoHunter:
@chebee7i You should turn on the option of "Build only if .travis.yml is present" option.

chebee7i:
You are a maintainer on it, so you should configure it however networkx-metis needs it to be configured.

OrkoHunter:
> You are a maintainer on it

I don't have the option to configure it (Travis?). I guess only owners and collaborators can do it.

chebee7i:
Oh sorry you meant for travis....yeah, I was referring to readthedocs.

OrkoHunter:
I edited the long line of this file. The line in `test_metis.py` is 80 characters, I wouldn't mind it. However the complete tests are to be re-written ( #26 ) so I guess things would be covered.

networkx/addons/metis/__init__.py
```diff
@@ -81,7 +81,7 @@ def function():
     def _convert_exceptions(func, *args, **kwargs):
         try:
             return func(*args, **kwargs)
-        except catch_types or () as e:
+        except (catch_types, ()) as e:
```
ysitu: Removing `or ()` should be good enough. `(catch_types, ())` is wrong.

OrkoHunter:
Are we good here?

PR #32, opened 2015-07-02
Move _metis.pyx and _api.pxd inside networkx.addons.metis namespace

OrkoHunter:


OrkoHunter:
If we all agree for this relocation, this can be merged. Currently we only have one `pyx` and `pxd` file, so it doesn't matter much, but if there are numerous such wrapper files, it would be potentially confusing as which `.pyx` is generating which `.so` and where. Henceforth, this structure will be followed.

setup.py
```diff
@@ -27,7 +27,8 @@
                'depends': glob('src/libmetis/*.h'),
                'include_dirs': ['src/GKlib', 'src/libmetis']})]
 
-ext_modules = cythonize([Extension('networkx.addons.metis._metis', ['*.pyx'],
+ext_modules = cythonize(
+  [Extension('networkx.addons.metis._metis', ['networkx/addons/metis/*.pyx'],
```
chebee7i: Shouldn't this be:  `[`networkx/addons/metis/_metis.pyx`]`

If knowing which pyx compiles to which extension module is important, then you should be explicit here.

OrkoHunter: Yeah correct. It is written implicitly here. Changing it.

PR #33, opened 2015-07-04
Add tests inside installed package

OrkoHunter:
We need this at first place for the tests to run inside `site-packages`.


chebee7i:
Why not just add an `__init__.py` file in the `tests` subdirectory?  How you have it currently is different from how we do it in networkx. Not saying that they necessary need to be the the same though...

OrkoHunter:
Having `package_data` for including tests (as it is in NetworkX) seems okay to me.

chebee7i:
It seems like putting them in `package_data` contradicts the fact that we have `__init__.py` files in the test directories. E.g., in networkx (and maybe here), we should just add the test directories to `packages`. This would simplify `package_data` quite a bit.

OrkoHunter:
Perfect. Added it into packages.

chebee7i:
I think this requires that you have an empty `__init__.py` in the `tests` subdirectory.

chebee7i:
Opps I'm blind.

OrkoHunter:
Thanks! I'll rebase and run the travis script.

PR #35, opened 2015-07-04
Add requirements.txt

### LINKED ISSUE #34: requirements.txt
It would be nice to have a requirements.txt for this project.  Cython needs to be included with in. Then installation from a cloned repo can be: `pip install -r requirements.txt`. I realize that setuptools takes care of most of the dependencies, but it does not catch Cython since we use it in `setup.py`.

OrkoHunter:
Fixes #34 

`networkx` needs to be `networkx>1.9.1`. But as we don't have a final decision on whether we are including it in `1.10` or not, I'm keeping it without the version number.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+Cython
+networkx
+decorator
+six
+enum34


chebee7i:
Yeah, we can always modify this later.

SanketDG:
Could someone fill me on this? Why are the rest of the packages here? This is duplication of information.
Only Cython and the test packages should be here, in case that is not handled by setup.py.

Also,can't Cython be handled with setup.py?

chebee7i:
We can certainly make setup.py pull this information from `requirements.txt`, but I think having a `requirements.txt` with Cython included is a good thing.

Cython cannot be handled with setup.py. At best, we can raise an `ImportError`, but what does that solve?

SanketDG:
> We can certainly make setup.py pull this information from requirements.txt, but I think having a requirements.txt with Cython included is a good thing.

Yes that could be done, but just having Cython is more explicit.

> Cython cannot be handled with setup.py. At best, we can raise an ImportError, but what does that solve?

Atleast it gives the user a heads up, instructions to install Cython is only written in the Readme. For a reference, we could just easily write that Python 2.7 isn't supported in the readme instead of checking for it in `setup.py`. Doing it in `setup.py` is way more intuitive instead of the Readme. 

Pandas does it: https://github.com/pydata/pandas/blob/master/setup.py#L80

OrkoHunter:
I must say that the given example of raising check on installation of Cython does not appropriately fit here. `NetworkX-Metis` is the repository where a Python and a C library is being interfaced. Raising a check saying "Cython not installed" sounds trivial to me. However, I agree that there is duplication of `install_requires` in setup.py and `requirements.txt`. But can't we consider both? After all, it's not much in the sense of code duplication.

SanketDG:
This is what happens when I dont have Cython:
```python
  File "setup.py", line 10, in <module>
    from Cython.Build import cythonize
ImportError: No module named 'Cython'
```
If this is helpful enough, then that's okay.

> However, I agree that there is duplication of install_requires in setup.py and requirements.txt. But can't we consider both? After all, it's not much in the sense of code duplication.

As you said, the difference would be trivial.

chebee7i:
>>We can certainly make setup.py pull this information from requirements.txt, but I think having a requirements.txt with Cython included is a good thing.

> Yes that could be done, but just having Cython is more explicit.

Isn't that somewhat counter to the point of a `requirements.txt`, which is to list all dependencies of the package?

All we have to do is something like:

with open('requirements.txt') as fobj:
    install_requires = fobj.readlines()

SanketDG:
> Isn't that somewhat counter to the point of a requirements.txt, which is to list all dependencies of the package?

For a application, yes. 
For a reusable library, its recommended to just use `setup.py`. We are using `pip` to distribute, so I cant reason on why we might need a requirements.txt.
http://blog.miguelgrinberg.com/post/the-package-dependency-blues

I think in this case, it can be perfectly used for listing dependencies that are not handled by setuptools i.e Cython and optionally, the test suite.

I now see too very clear options:

- Read all dependencies from requirements.txt into `install_requires` and atleast include an exception for a Cython `ImportError` (I do not find this trivial)

- Keep setup.py as it is, include Cython and the test suite in requirements.txt and then instruct users to install from there after runnning setup.py

SanketDG:
I found another pretty way of handling it, they just include the compiled .c file, lines 50 - 56: http://bazaar.launchpad.net/~jerub/oursql/master/view/head:/setup.py

chebee7i:
Aha yes, that is also a pretty good option. I think NumPy does (or did) that too. So we could remove the Cython dependency altogether. Then the requirements.txt is mostly a convenience for developers and would only include what `install_requires`does not.

PR #36, opened 2015-07-08
Drop namespace packaging

OrkoHunter:
This PR removes all the dependency of `networkx-metis` on the namespace packaging and will live inside `networx_metis-xxx.egg` as a normal package.

This will obviously create two ways of using the library
```py
from networkx.addons import metis
metis.partition()
```
and
```py
import networkx_metis
networkx_metis.partition()
```

So, it's all upon the user to choose between the two methods. Kind of confusing though :(

Regarding `.travis.yml`, I've changed it to simply `nosetests metis.tests.test_metis`. We can easily change it when related changes are done in NetworkX core.


ysitu:
Shall we make it concise and use `nxmetis`? A number of scikits packages are named `skxxx`.

OrkoHunter:
I've been thinking for that too. In the beginning, when I used to test the package on my machine, I preferred naming the package as `nxmetis`.

ysitu:
In `.travis.yml`, you do not need install a custom version of NetworkX anymore.

chebee7i:
Why will it still be possible to do: `from networkx.addons import metis`?  I was assuming `networkx.addons` would not exist at all.

OrkoHunter:
> Flask extensions should urge users to import from flask.ext.foo instead of flask_foo or flaskext_foo so that extensions can transition to the new package name without affecting users.

It's upon us how we want to present this addon.

.travis.yml
```diff
@@ -15,12 +15,12 @@ before_install:
 
 install:
  - pip install Cython
- - pip install https://github.com/networkx/networkx/archive/master.zip
+ - pip install networkx
```
ysitu: You do not need this.

nxmetis/__init__.py
```diff
@@ -15,10 +15,10 @@
...
 import six
 
-from networkx.addons.metis import enums
-from networkx.addons.metis import exceptions
-from networkx.addons.metis import _metis
-from networkx.addons.metis import types
+from . import enums
+from . import exceptions
+from . import _metis
+from . import types
```
ysitu: Use absolute imports.

nxmetis/_metis.pyx
```diff
@@ -7,7 +7,7 @@ import os
 import sys
 import tempfile
 
-from networkx.addons.metis import exceptions
+from . import exceptions
```
ysitu: Use absolute imports.

nxmetis/exceptions.py
```diff
@@ -1,4 +1,4 @@
-from networkx.addons.metis import enums
+from . import enums
```
ysitu: Use absolute imports.

nxmetis/tests/test_metis.py
```diff
@@ -1,9 +1,9 @@
 import itertools
 import nose.tools
 
-from networkx.addons.metis import exceptions
-from networkx.addons.metis import _metis
-from networkx.addons.metis import types
+from .. import exceptions
+from .. import _metis
+from .. import types
```
ysitu: Use absolute imports.

nxmetis/types.py
```diff
@@ -1,7 +1,7 @@
 import numbers
 
-from networkx.addons.metis import enums
-from networkx.addons.metis import _metis
+from . import enums
+from . import _metis
```
ysitu: Use absolute imports.

setup.py
```diff
@@ -28,8 +28,8 @@
                'include_dirs': ['src/GKlib', 'src/libmetis']})]
 
 ext_modules = cythonize(
-    [Extension('networkx.addons.metis._metis',
-               ['networkx/addons/metis/_metis.pyx'],
+    [Extension('nxmetis._metis',
+               ['nxmetis/_metis.pyx'],
                include_dirs=['src/GKlib', 'src/libmetis'],
                libraries=['metis', 'gklib'])])
```
ysitu: Make this more compact.

setup.py
```diff
@@ -59,11 +59,8 @@
                               'Programming Language :: Python :: 3.2',
                               'Programming Language :: Python :: 3.3',
                               'Programming Language :: Python :: 3.4',],
-        packages           = ['networkx',
-                              'networkx.addons',
-                              'networkx.addons.metis',
-                              'networkx.addons.metis.tests'],
-        namespace_packages = ['networkx.addons', 'networkx'],
+        packages           = ['nxmetis',
+                              'nxmetis.tests'],
```
ysitu: Make this more compact.

PR #39, opened 2015-07-10
Add appveyor.yml and necessary files to build tests on AppVeyor

OrkoHunter:
This is more or less how sckit-learn does it. I've removed unnecessary parts and we can do more but first we need to setup the hook on https://ci.appveyor.com to start the first build. @chebee7i Can you setup the hook for networkx-metis?


appveyor.yml
```diff
@@ -0,0 +1,56 @@
...
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+
+  # Install the build and runtime dependencies of the project.
+  - "%CMD_IN_ENV% pip install --timeout=60 -r requirements.txt"
+  - "python setup.py build --compiler=msvc"
+  - "python setup.py install"
+
+# Not a .NET project, we build scikit-learn in the install step instead
+build: false
```
MridulS: `scikit-learn`

appveyor/run_with_env.cmd
```diff
@@ -0,0 +1,47 @@
...
+    SET MSSdk=1
+    "%WIN_SDK_ROOT%\%WINDOWS_SDK_VERSION%\Setup\WindowsSdkVer.exe" -q -version:%WINDOWS_SDK_VERSION%
+    "%WIN_SDK_ROOT%\%WINDOWS_SDK_VERSION%\Bin\SetEnv.cmd" /x64 /release
+    ECHO Executing: %COMMAND_TO_RUN%
+    call %COMMAND_TO_RUN% || EXIT 1
+) ELSE (
+    ECHO Using default MSVC build environment for 32 bit architecture
+    ECHO Executing: %COMMAND_TO_RUN%
+    call %COMMAND_TO_RUN% || EXIT 1
+)
```
MridulS: newline

MridulS:
@OrkoHunter did you build this on appveyor. This doesn't seem to work for me.

OrkoHunter:
Yes, there's a C library build problem. `msvc` seems to be throwing lots of build failures which `gcc` as a compiler have been accepting. Here's [the build log of my fork](https://ci.appveyor.com/project/OrkoHunter/networkx-metis/build/1.0.19/job/9sb7fhdjwxk6787q).

ysitu:
The problem is in https://github.com/networkx/networkx-metis/blob/master/src/GKlib/gk_types.h#L22. `ssize_t` is not defined in MSVC. You need to add something like

```c
#ifdef _MSC_VER
#include <basetsd.h>
typedef SSIZE_t ssize_t;
#endif
```

appveyor.yml
```diff
@@ -0,0 +1,56 @@
...
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "64"
+
+    - PYTHON: "C:\\Python34"
+      PYTHON_VERSION: "3.4.1"
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python34-x64"
+      PYTHON_VERSION: "3.4.1"
+      PYTHON_ARCH: "64"
```
ysitu: According to http://www.appveyor.com/docs/installed-software, installed Python versions are 2.7.9, 3.3.5 and 3.4.3 for both x86 and x64.

appveyor.yml
```diff
@@ -0,0 +1,56 @@
...
+  - "python setup.py install"
+
+# Not a .NET project, we build scikit-learn in the install step instead
+build: false
+
+test_script:
+  # Change to a non-source folder to make sure we run the tests on the
+  # installed library.
+  - "mkdir empty_folder"
+  - "cd empty_folder"
```
ysitu:
```cmd
cd %TEMP%
```

probably suffices.

appveyor/install.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+    # Download and retry up to 3 times in case of network transient errors.
+    Write-Host "Downloading" $filename "from" $url
+    $retry_attempts = 2
+    for($i=0; $i -lt $retry_attempts; $i++){
+        try {
+            $webclient.DownloadFile($url, $filepath)
+            break
+        }
+        Catch [Exception]{
```
ysitu: Add space before `{`.

appveyor.yml
```diff
@@ -0,0 +1,56 @@
...
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python34-x64"
+      PYTHON_VERSION: "3.4.1"
+      PYTHON_ARCH: "64"
+
+install:
+  # Install Python (from the official .msi of http://python.org) and pip when
+  # not already installed.
+  - "powershell ./appveyor/install.ps1"
```
ysitu: Use backslashes.

chebee7i:
`networkx-metis` should be ready for commits now.

appveyor.yml
OrkoHunter: Probably not here. I used backslashes and there was
```
Error parsing appveyor.yml: (Line: 40, Col: 5, Idx: 1079) - (Line: 40, Col: 27, Idx: 1101): While parsing a quoted scalar, find unknown escape character.
```

OrkoHunter:
Container-based Travis environment has become really fast.

ysitu:
@OrkoHunter Will you help port the Travis CI changes to `networkx`?

OrkoHunter:
Yes, I'm sending a PR shortly.

appveyor.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.9"
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.9"
+      PYTHON_ARCH: "64"
+
```
ysitu: Include Python 3.2.6 since you have declared support for that in `setup.py`. I think that you can simply add it here without doing anything extra. `install.ps1` should handle the installation.

MridulS:
@OrkoHunter Are you also working on making AppVeyor testing for networkx?

OrkoHunter:
> @OrkoHunter Are you also working on making AppVeyor testing for networkx?

As soon as it gets working with `networkx-metis`, it can be ported to `networkx`.

OrkoHunter:
The latest error on the build is 
```
src/GKlib\gk_types.h(24) : error C2371: 'ssize_t' : redefinition; different basic types 
        c:\projects\networkx-metis\src\gklib\gk_arch.h(53) : see declaration of 'ssize_t' 
```
https://ci.appveyor.com/project/OrkoHunter/networkx-metis/build/1.0.27/job/h9lgxxsgahc27gwf#L399

ysitu:
Like I said, after changing `WIN32` to `_WIN32`, you should be able to revert the change to `gk_types.h`.

OrkoHunter:
Ah! Sorry.

appveyor.yml
ysitu: Hmm... 3.2.6 is a source-only release. The latest binary release is 3.2.5.

src/GKlib/timers.c
```diff
@@ -7,7 +7,6 @@
 \version\verbatim $Id: timers.c 10711 2011-08-31 22:23:04Z karypis $ \endverbatim
 */
 
-
```
ysitu: Avoid gratuitous changes.

appveyor.yml
OrkoHunter: Okay.. The library was built successfully. Strange that it couldn't find `nose`. https://ci.appveyor.com/project/OrkoHunter/networkx-metis/build/1.0.28/job/2w5r4cqq45tvh562#L2312

I added the command to install `nose`. But it is installed already.

OrkoHunter:
@ysitu Can you cancel all the builds except for the latest one? https://ci.appveyor.com/project/chebee7i/networkx-metis-1lf0f/history
I think @chebee7i could do it only because it's under `project/chebee7i`

appveyor.yml
```diff
@@ -0,0 +1,72 @@
...
+  - "pip install nose"
+
+# Not a .NET project, we build networkx-metis in the install step instead
+build: false
+
+test_script:
+  # Change to a non-source folder to make sure we run the tests on the
+  # installed library.
+  - "cd %TEMP%"
+  - "python -c \"import nose; nose.main()\" -s -v nxmetis"
```
ysitu: Use

```python
nose.main(module='nxmetis.tests.test_metis')
```

See http://nose.readthedocs.org/en/latest/api/core.html.

ysitu:
Apparently @chebee7i owns the CI as you said, so I can do nothing.

appveyor.yml
```diff
@@ -0,0 +1,72 @@
...
+  # Check that we have the expected version and architecture for Python
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+
+  # Install the build and runtime dependencies of the project.
+  - "%CMD_IN_ENV% pip install --timeout=60 -r requirements.txt"
+  - "python setup.py build --compiler=msvc"
+  - "python setup.py install"
+
+  - "pip install nose"
```
ysitu: Do you need `%CMD_IN_ENV%` when running `pip`?

OrkoHunter: Yes.. `%CMD_IN_ENV%` is used while running `pip`, `conda`, etc.

ysitu: Perhaps only for stuff that requires builds extensions? Then it makes sense that you don't use it on this line.

appveyor.yml
OrkoHunter: This is still not being able to find `nxmetis` from `%TMP%`. Works perfectly from the local directory.

OrkoHunter: Only for Python27 though. It's finding the package in Python32.

OrkoHunter:
@chebee7i if you will cancel all the builds except for the last one, probably 7-8 hours of time can be saved :)

chebee7i:
Hmm...unfortunately, it doesn't appear that `appveyor` allows me to give others access (as readthedocs does).  I will go ahead and cancel, but we should try to figure out a better solution for "organizations".

chebee7i:
It might be worth emailing them about it. I will be travelling tomorrow for most of the day, so I expect to be unavailable.

chebee7i:
I cancelled all builds, please initiate a new build.

OrkoHunter:
Thanks @chebee7i. Yeah, I really liked how readthedocs gives the preferences option. I'll try opening an issue.

appveyor.yml
```diff
@@ -0,0 +1,72 @@
...
+build: false
+
+test_script:
+  # Change to a non-source folder to make sure we run the tests on the
+  # installed library.
+  - "cd %TEMP%"
+  - "nosetests -v nxmetis.tests.test_metis"
+
+  # Move back to the project folder
+  - "cd .."
```
ysitu: This does not work. You are going into `%TEMP%\..`. Apparently you can use `pushd` and `popd` even on Windows: https://en.wikipedia.org/wiki/Pushd_and_popd.

ysitu:
This is looking good. Please update `NOTICE` to include the additional changes you have done to METIS and also credit the authors of `install.ps1` and `run_with_env.cmd` there.

OrkoHunter:
The test for `part_graph` function failed on AppVeyor. It's giving out a different answer than expected. Also, it's strange that test passes on Travis.

ysitu:
Well, METIS does not guarantee that it will always generate the optimal solution since it is an approximation algorithm to NP-hard problems. We need to allow some variation in the test output.

OrkoHunter:
Okay then. I'll update the `NOTICE`.

OrkoHunter:
Any more changes in the `NOTICE` or correction?

ysitu:
That looks good to me. Let's fix the test in a separate PR.

Issue #41, opened 2015-07-15
Fix AppVeyor CI


ysitu:
This is a requirement for declaring Windows support.

OrkoHunter:
Rewriting the tests #26 would do it.

OrkoHunter:
There are two failures over the Appveyor CI in [the latest build](https://ci.appveyor.com/project/chebee7i/networkx-metis-1lf0f/build/1.0.37) only because of the x64 architecture. There seems to be a bug probably unresolved for compiling `msvc9compiler.py` over x64 windows machines. It has temporary fixes like changing `vcvarsall.bat` file somehow.

OrkoHunter:
I think it can be tricky to do the fix over Appveyor. Maybe we do not test for x64 ?

PR #44, opened 2015-07-16
Rewrite test_metis.py to test exposed functions

### LINKED ISSUE #26: Change test_metis.py to test the exposed functions

OrkoHunter:
Fixes #26 

I'm not sure whether it's the best testing method but I've used a `nx.complete_graph(16)` and tested the results against what I got on my computer.

As the results are generated by algorithms based on approximations, they depend upon the platform and compiler used.

### DIFF
--- a/nxmetis/tests/test_metis.py
+++ b/nxmetis/tests/test_metis.py
@@ -1,6 +1,9 @@
 import itertools
 import nose.tools
 
+import networkx as nx
+
+import nxmetis
 from nxmetis import exceptions
 from nxmetis import _metis
 from nxmetis import types
@@ -17,59 +20,81 @@ def make_cycle(n):
 
 class TestMetis(object):
 
-    def test_node_nd(self):
-        n = 16
-        xadj, adjncy = make_cycle(n)
-        perm, iperm = _metis.node_nd(xadj, adjncy)
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def setUp(self):
+        self.node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
+                          1, 2, 3, 4, 5, 6]
+        self.G = nx.Graph()
+        self.G.add_path(self.node_list)
+        self.G.add_edge(self.node_list[-1], self.node_list[0])
 
-    def test_selfloops(self):
-        n = 16
-        xadj = list(range(0, 3 * n + 1, 3))
-        adjncy = list(
-            itertools.chain.from_iterable(
-                zip(itertools.chain([n - 1], range(n - 1)),
-                    range(n),
-                    itertools.chain(range(1, n), [0]))))
-        perm, iperm = _metis.node_nd(xadj, adjncy)
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def test_node_nested_dissection_unweighted(self):
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.assert_equal(len(self.G), len(node_ordering))
+        nose.tools.assert_equal(set(self.G), set(node_ordering))
 
-    def test_part_graph(self):
-        n = 16
-        xadj, adjncy = make_cycle(n)
-        for recursive in (False, True):
-            objval, part = _metis.part_graph(
-                xadj, adjncy, 2, recursive=recursive)
-            nose.tools.assert_equal(objval, 2)
-            nose.tools.assert_equal(set(part), set(range(2)))
-            it = itertools.dropwhile(lambda x: x == 0, itertools.cycle(part))
-            nose.tools.assert_equal(
-                list(itertools.takewhile(lambda x: x == 1, it)), [1] * (n // 2))
+        # Tests for exercising package's ability to handle self-loops
+        # METIS crashes on self loops. networkx-metis should not
+        self.G.add_edge(1, 1)
+        self.G.add_edge('a', 'a')
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.assert_equal(len(self.G), len(node_ordering))
+        nose.tools.assert_equal(set(self.G), set(node_ordering))
 
-    def test_compute_vertex_separator(self):
-        n = 16
-        xadj, adjncy = make_cycle(n)
-        sepsize, part = _metis.compute_vertex_separator(xadj, adjncy)
-        nose.tools.assert_equal(sepsize, 2)
-        nose.tools.assert_equal(len(part), n)
-        part1, part2, sep = (list(filter(lambda i: part[i] == k, range(n)))
-                             for k in range(3))
-        nose.tools.assert_equal(sorted(part1 + part2 + sep), list(range(n)))
+    def test_partition(self):
+        partition = nxmetis.partition(self.G, 4)
+        # When we choose one node from one part of the partitioned Graph,
+        # It must be adjacent to one or more of the nodes in the same part.
+        # This is to verify the continuity of the chain of nodes.
+        parts = partition[1]  # List containing partitioned node lists
+
+        nose.tools.assert_equal(list(partition)[0], 4)
+
+        for part in parts:
+            nose.tools.assert_not_equal(0, len(part))  # Non-empty set
+            nose.tools.assert_equal(len(part), len(set(part)))  # Duplicate-free
+            nose.tools.ok_(nx.is_connected(self.G.subgraph(part)))  # Connected
+
+        # Disjoint sets
+        for part1, part2 in itertools.combinations(parts, 2):
+            nose.tools.assert_equal(set(), set(part1) & set(part2))
+
+        # These parts must be exhaustive with the node list of the Graph
+        parts_combined = parts[0] + parts[1] + parts[2] + parts[3]
+        nose.tools.assert_equal(set(parts_combined), set(self.G))
+
+    def test_vertex_separator(self):
+        bisection = nxmetis.vertex_separator(self.G)
+        sep, part1, part2 = nxmetis.vertex_separator(self.G)
+
+        # The two separator nodes must not be present in the
+        # two bisected chains
+        nose.tools.ok_(sep[0] not in part1)
+        nose.tools.ok_(sep[0] not in part2)
+        nose.tools.ok_(sep[1] not in part1)
+        nose.tools.ok_(sep[1] not in part2)
+
+        # There should be two different separator nodes
         nose.tools.assert_equal(len(sep), 2)
-        nose.tools.assert_equal(abs(sep[1] - sep[0]), n // 2)
-        nose.tools.assert_equal(
-            sorted(map(sorted, [part1, part2])),
-            sorted(map(sorted,
-                       [[(sep[0] + i) % n for i in range(1, n // 2)],
-                        [(sep[1] + i) % n for i in range(1, n // 2)]])))
+        nose.tools.assert_not_equal(sep[0], sep[1])
+
+        # The lists should be exhaustive with the node list of the Graph
+        nose.tools.assert_equal(set(sep) | set(part1) | set(part2),
+                                set(self.G))
+
+        # The parts must be disjoint sets
+        nose.tools.assert_equal(set(), set(part1) & set(part2))
+
+        # Non-empty set
+        nose.tools.assert_not_equal(len(part1), 0)
+        nose.tools.assert_not_equal(len(part2), 0)
+
+        # Duplicate-free
+        nose.tools.assert_equal(len(part1), len(set(part1)))
+        nose.tools.assert_equal(len(part2), len(set(part2)))
+
+        # Connected
+        nose.tools.ok_(nx.is_connected(self.G.subgraph(part1)))
+        nose.tools.ok_(nx.is_connected(self.G.subgraph(part2)))
 
     def test_MetisOptions(self):
         n = 16


OrkoHunter:
For `nxmetis.partition(G, 4)`
```
with msvc on Windows - (96, [[1, 2, 3, 9], [0, 4, 5, 6], [7, 8, 10, 15], [11, 12, 13, 14]])
with gcc on Linux   -  (96, [[1, 2, 3, 10], [4, 5, 6, 7], [8, 9, 12, 14], [0, 11, 13, 15]])
```

OrkoHunter:
It appears that we are not the only one facing this problem. 

http://stackoverflow.com/questions/31299159/metis-different-results-on-different-os
http://stackoverflow.com/questions/29397370/getting-non-unique-results-from-metis-when-different-compiler-is-used

OrkoHunter:
I think the good way here to test METIS should be that if Metis returns couples of nested lists or tuple (the way we except) the tests should *pass* unless there is some error returned. Eventually we can't be sure for what METIS is going to return for a particular operation.

nxmetis/tests/test_metis.py
```diff
@@ -1,10 +1,14 @@
...
 import nose.tools
 
+import networkx as nx
+
+import nxmetis
 from nxmetis import exceptions
 from nxmetis import _metis
 from nxmetis import types
 
+G = nx.complete_graph(16)  # An unweighted complete Graph made of 16 nodes
```
ysitu: Move this into to `setUp()` function under `TestMetis`:

```python
def setUp(self):
    self.G = nx.complete_graph(16)
```

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +21,18 @@ def make_cycle(n):
...
-        perm, iperm = _metis.node_nd(xadj, adjncy)
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def test_node_nd_unweighted(self):
+        node_order = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+        node_ordering = nxmetis.node_nested_dissection(G)
+        nose.tools.assert_equal(node_ordering, node_ordering)
```
ysitu: What does this test? Do you mean `assert_equal(sorted(G), sorted(node_ordering))` (or `six.assertCountEqual(self, G, node_ordering)`)?

ysitu:
It is true that we should tolerate variation in METIS output. But we should still verify the invariants. For example, when we do graph partitioning, we should verify that the returned partitions do not overlap.

nxmetis/tests/test_metis.py
ysitu: Also, consider using a graph whose nodes are not integers. That way we can verify that we are returning nodes to users instead of their indices in enumeration.

nxmetis/tests/test_metis.py
OrkoHunter: Ooops.. I forgot to change the names.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def setUp(self):
+        node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
+                     1, 2, 3, 4, 5, 6]
+        self.G = nx.Graph()
+        self.G.add_edges_from(list(itertools.combinations(node_list, 2)))
```
ysitu: I suggest that you use a cycle graph. That gives you some more interesting invariants to check. For example, when you partition it, you should expect to get some contiguous chains of nodes.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
 
-    def test_selfloops(self):
```
ysitu: Please add a test that exercises the package's ability to handle self loops. IIRC, METIS crashes when you have self loops in graphs.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
-                zip(itertools.chain([n - 1], range(n - 1)),
-                    range(n),
-                    itertools.chain(range(1, n), [0]))))
-        perm, iperm = _metis.node_nd(xadj, adjncy)
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def test_node_nd_unweighted(self):
```
ysitu: `test_node_nested_dissection_unweighted`

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.assert_equal(sorted(self.G), sorted(node_ordering))
 
     def test_part_graph(self):
```
ysitu: `test_partition`

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
-        n = 16
-        xadj, adjncy = make_cycle(n)
-        for recursive in (False, True):
```
ysitu: Please keep this loops as the underlying METIS functions are different.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
-            nose.tools.assert_equal(objval, 2)
-            nose.tools.assert_equal(set(part), set(range(2)))
-            it = itertools.dropwhile(lambda x: x == 0, itertools.cycle(part))
-            nose.tools.assert_equal(
-                list(itertools.takewhile(lambda x: x == 1, it)), [1] * (n // 2))
+        partitioning = (96, [[1, 'c', 'b', 'j'], ['e', 'd', 'g', 'f'],
+                        ['i', 'h', 3, 6], ['a', 2, 4, 5]])
+        nose.tools.assert_equal(nxmetis.partition(self.G, 4), partitioning)
 
     def test_compute_vertex_separator(self):
```
ysitu: `test_vertex_separator`

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,25 @@ def make_cycle(n):
...
-        nose.tools.assert_equal(len(sep), 2)
-        nose.tools.assert_equal(abs(sep[1] - sep[0]), n // 2)
-        nose.tools.assert_equal(
-            sorted(map(sorted, [part1, part2])),
-            sorted(map(sorted,
-                       [[(sep[0] + i) % n for i in range(1, n // 2)],
-                        [(sep[1] + i) % n for i in range(1, n // 2)]])))
+        bisection = ([1, 'c', 'e', 'g', 'f', 'h', 4],
+                     ['a', 'b', 'd', 'i', 'j', 2, 3, 6, 5], [])
+        nose.tools.assert_equal(nxmetis.vertex_separator(self.G), bisection)
```
ysitu: Using a cycle graph will allow you have a nonempty third component in the output.

nxmetis/tests/test_metis.py
```diff
@@ -14,18 +18,46 @@ def make_cycle(n):
...
+
+    def test_partition(self):
+        partitioning = (4, [[1, 2, 3, 4], ['a', 'b', 5, 6],
+                        ['c', 'd', 'e', 'f'], ['g', 'h', 'i', 'j']])
+        i = list(partitioning)[1]
+        partition = nxmetis.partition(self.G, 4)
+        j = list(partition)
+        nose.tools.assert_true(seq_in_seq(list(partition)[1][0], self.node_list*2))
+#       nose.tools.assert_true(seq_in_seq(list(partition)[1][1], self.node_list*2))
+#       ['a', 'b', 5, 6] ??
```
OrkoHunter: TODO

OrkoHunter:
I made some changes regarding the *test-the-continuous-chains-from-partition*. I'll try to finish it later today.

nxmetis/tests/test_metis.py
```diff
@@ -14,18 +18,46 @@ def make_cycle(n):
...
-        perm, iperm = _metis.node_nd(xadj, adjncy)
-        nose.tools.assert_equal(set(perm), set(range(n)))
-        nose.tools.assert_equal(abs(perm[-1] - perm[-2]), n // 2)
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def setUp(self):
+        self.node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
+                     1, 2, 3, 4, 5, 6]
+        class OrderedGraph(nx.Graph):
```
ysitu: I do not think that using an ordered graph helps. We do not know how METIS handles node ordering internally.

OrkoHunter: The results produced by METIS varies across different platforms even for the same input. Using and OrderedGraph helps us to detect the "chain" which must be the output of partitioning a cycle graph. This is sure a narrow pathway.

ysitu: Let's consider what we can (reasonably) do if we have a cycle graph of 16 nodes whose identities we do not even know.

* `node_nested_dissection`: This function is about finding a node elimination order that reduces fill (think efficient sparse Cholesky factorization). We cannot do much other than verifying `set{G} == set{result}`.
* `partition`: Let's partition into say four parts. It is reasonable to assume, despite without guarantee, that METIS will return four consecutive segments. We can verify that each of those are nonempty, contiguous and disjoint, and their union contains the same nodes as `G`.
* `vertex_separator`: Again, it is reasonable to assume in spite of lack of guarantee that METIS will return a separator of exactly two nodes with the remaining nodes forming two contiguous chains. If we consider the two separator nodes two singleton chains, we can verify their correctness the same way we do for `partition` (perhaps ignoring the nonemptiness condition).

OrkoHunter: So far, I think we are good with testing `node_nested_dissection `. We cannot do much than verifying that they generate equal sets.

Regarding `partition `,  now I think we can verify the continuity without the use of OrderedGraph. After all we just have to know the order in `G.nodes()`.

Thanks for the suggestion regarding `vertex_separator`. I'll try to implement it very soon.

nxmetis/tests/test_metis.py
```diff
@@ -14,18 +18,56 @@ def make_cycle(n):
...
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+    def setUp(self):
+        self.node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
+                     1, 2, 3, 4, 5, 6]
+        self.G = nx.Graph()
+        edges = [(self.node_list[i], self.node_list[i+1])
+                 for i in range(len(self.node_list) - 1)]
+        edges.append((6, 'a'))  # To complete the cycle
```
ysitu:
```python
G.add_path(node_list)
G.add_edge(node_list[-1], node_list[0])
```

nxmetis/tests/test_metis.py
```diff
@@ -42,34 +84,12 @@ def test_selfloops(self):
                        (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
         nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
```
ysitu: You can simplify this by simply adding selfloops to add nodes in `self.G`. Use `node_nested_dissection` for testing.

OrkoHunter: Did you mean something like [this](https://github.com/networkx/networkx-metis/pull/44/files#diff-fbfe1ddde32d89fa89c2191c21aa9d6dR86)?

appveyor.yml
```diff
@@ -68,5 +68,3 @@ test_script:
   - "pushd %TEMP%"
   - "nosetests -v nxmetis.tests.test_metis"
 
-  # Move back to the project folder
-  - "popd"
```
OrkoHunter: This was causing the build failure even when the tests would pass. Piggyback it here or a new PR/commit?

OrkoHunter: I think it should be in one of the fixes for #41

OrkoHunter:
The new tests are ready for a review. Please ignore the lines commented. I'll remove the useless code before merging.

nxmetis/tests/test_metis.py
```diff
@@ -14,19 +18,60 @@ def make_cycle(n):
...
+    def test_node_nested_dissection_unweighted(self):
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.assert_equal(set(self.node_list).intersection(node_ordering),
+                               set(self.node_list))
+        nose.tools.assert_equal(set(self.node_list).intersection(node_ordering),
+                               set(node_ordering))
+
+    def test_partition(self):
+        partitioning = (4, [[3, 4, 6, 5], ['a', 'c', 'b', 'd'],
+                        ['e', 'g', 'f', 'h'], [1, 'i', 'j', 2]])
```
ysitu: You should not assume that a specific partitioning will be returned.

OrkoHunter: Yes, it's not assumed. `partitioning` is not used anywhere. I'll remove it.

OrkoHunter: Let me remove all the useless code from here.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,65 @@ def make_cycle(n):
...
+        self.G = nx.Graph()
+        self.G.add_path(self.node_list)
+        self.G.add_edge(self.node_list[-1], self.node_list[0])
+
+    def test_node_nested_dissection_unweighted(self):
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.ok_(set(self.node_list) & set(node_ordering) ==
+                       set(self.node_list))
+        nose.tools.ok_(set(self.node_list) & set(node_ordering) ==
+                       set(node_ordering))
```
ysitu: Did you notice that you are checking just `set(node_list) == set(node_ordering)` in a very convoluted way? You are also missing the case where there may be duplicates in `node_ordering`. You can do this instead:

```python
assert_equal(len(G), len(node_ordering))
assert_equal(set(G), set(node_ordering))
```

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,65 @@ def make_cycle(n):
...
+        # When we choose one node from one part of the partitioned Graph,
+        # It must be adjacent to one or more of the nodes in the same part.
+        # This is to verify the continuity of the chain of nodes.
+        parts = list(partition)[1]  # List containing partitioned node lists
+        for i in range(4):
+            for j in range(4):
+                nose.tools.assert_true(parts[i][j] in self.G[parts[i][0]] or
+                                       parts[i][j] in self.G[parts[i][1]] or
+                                       parts[i][j] in self.G[parts[i][2]] or
+                                       parts[i][j] in self.G[parts[i][3]])
```
ysitu: You are not checking whether `partition` returns `4` in the first component. Neither are you checking whether the partitions are nonempty, duplicate-free and disjoint. Finally, to verify connectedness, it is much simpler to use

```python
for part in parts:
    nose.tools.ok_(nx.is_connected(G.subgraph(part)))
```

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,65 @@ def make_cycle(n):
...
-        nose.tools.ok_(set(range(min(perm[-2:]) + 1, max(perm[-2:]))) in
-                       (set(perm[0:n // 2 - 1]), set(perm[n // 2 - 1:-2])))
-        nose.tools.ok_(all(i == perm[iperm[i]] for i in range(n)))
+        self.G.add_edge(1, 1)
+        self.G.add_edge('a', 'a')
+        node_ordering = nxmetis.node_nested_dissection(self.G)
+        nose.tools.ok_(set(self.node_list) & set(node_ordering) ==
+                       set(self.node_list))
+        nose.tools.ok_(set(self.node_list) & set(node_ordering) ==
+                       set(node_ordering))
```
ysitu: How about merge this test with `test_node_nested_dissection_unweighted` and add a comment about the purpose of the selfloops?

OrkoHunter: Yeah, it seemed verbose to me too.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,65 @@ def make_cycle(n):
...
-        xadj, adjncy = make_cycle(n)
-        for recursive in (False, True):
-            objval, part = _metis.part_graph(
-                xadj, adjncy, 2, recursive=recursive)
-            nose.tools.assert_equal(objval, 2)
-            nose.tools.assert_equal(set(part), set(range(2)))
-            it = itertools.dropwhile(lambda x: x == 0, itertools.cycle(part))
-            nose.tools.assert_equal(
-                list(itertools.takewhile(lambda x: x == 1, it)), [1] * (n // 2))
+    def test_vertex_separator(self):
```
ysitu: Simplify this test the same way as `test_partition`.

nxmetis/tests/test_metis.py
OrkoHunter: Aah! I don't know what was I thinking at that time. Changing it.

nxmetis/tests/test_metis.py
OrkoHunter: Yes, it is much simpler to use `is_connected` with `subgraph` . Thank you.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,75 @@ def make_cycle(n):
...
+        nose.tools.assert_equal(len(self.G), len(node_ordering))
+        nose.tools.assert_equal(set(self.G), set(node_ordering))
+
+
+    def test_partition(self):
+        partition = nxmetis.partition(self.G, 4)
+        # When we choose one node from one part of the partitioned Graph,
+        # It must be adjacent to one or more of the nodes in the same part.
+        # This is to verify the continuity of the chain of nodes.
+        parts = list(partition)[1]  # List containing partitioned node lists
```
ysitu: `partition` is a tuple. You should not need `list()`.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,75 @@ def make_cycle(n):
...
+
+        nose.tools.assert_equal(list(partition)[0], 4)
+
+        for part in parts:
+            nose.tools.ok_(set(part)!={0})  # Non-empty set
```
ysitu: Should be `len(part) != 0`. Add spaces before and after `!=`.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,75 @@ def make_cycle(n):
...
+            nose.tools.ok_(len(part)==len(set(part)))  # Duplicate-free
```
ysitu: Use `assert_equal()`.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,75 @@ def make_cycle(n):
...
+            nose.tools.ok_(nx.is_connected(self.G.subgraph(part)))  # Connected
+        
+        # Disjoint sets
+        nose.tools.ok_(set(parts[0]) & set(parts[1]) == set())
+        nose.tools.ok_(set(parts[0]) & set(parts[2]) == set())
+        nose.tools.ok_(set(parts[0]) & set(parts[3]) == set())
+        nose.tools.ok_(set(parts[1]) & set(parts[2]) == set())
+        nose.tools.ok_(set(parts[1]) & set(parts[3]) == set())
+        nose.tools.ok_(set(parts[2]) & set(parts[3]) == set())
```
ysitu:
```python
for part1, part2 in itertools.combinations(parts, 2):
    assert_equal(set(), set(part1) & set(part2))
```

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,75 @@ def make_cycle(n):
...
+
+    def test_vertex_separator(self):
+        bisection = nxmetis.vertex_separator(self.G)
+
+        # The two separator nodes must not be present in the
+        # two bisected chains
+        nose.tools.ok_(list(bisection)[0][0] not in list(bisection)[1])
+        nose.tools.ok_(list(bisection)[0][0] not in list(bisection)[2])
+        nose.tools.ok_(list(bisection)[0][1] not in list(bisection)[1])
+        nose.tools.ok_(list(bisection)[0][1] not in list(bisection)[2])
```
ysitu: This is getting hard to read. Use

```python
sep, part1, part2 = nxmetis.vertex_separator(self.G)
```

and refer to the three return values by name.

OrkoHunter:
Last three comments addressed.

OrkoHunter:
I've made the changes and rebased it because of #47. We can wait for a green tick on appveyor but that might take long.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,70 @@ def make_cycle(n):
...
+            nose.tools.assert_equal(set(), set(part1) & set(part2))
+    def test_vertex_separator(self):
+        bisection = nxmetis.vertex_separator(self.G)
+        sep, part1, part2 = nxmetis.vertex_separator(self.G)
+        # The two separator nodes must not be present in the
+        # two bisected chains
+        nose.tools.ok_(sep[0] not in part1)
+        nose.tools.ok_(sep[0] not in part2)
+        nose.tools.ok_(sep[1] not in part1)
+        nose.tools.ok_(sep[1] not in part2)
```
ysitu: You need to first check that `len(sep)` is two. You also need to verify that `sep[0]` is different from `seq[1]`.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,70 @@ def make_cycle(n):
...
+        
+        # The lists should be exhaustive with the node list of the Graph
+        nose.tools.ok_(set(sep) | set(part1) |
+                       set(part2) == set(self.G))
+
+        # Continuity check
+        part1 = part1  # One of the two bisected lists
+        part2 = part2  # One of the two bisected lists
```
ysitu: Remove these two lines.

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,70 @@ def make_cycle(n):
...
+        nose.tools.assert_equal(list(partition)[0], 4)
+
+        for part in parts:
+            nose.tools.assert_not_equal(0, len(part))  # Non-empty set
+            nose.tools.assert_equal(len(part), len(set(part)))  # Duplicate-free
+            nose.tools.ok_(nx.is_connected(self.G.subgraph(part)))  # Connected
+        
+        # Disjoint sets
+        for part1, part2 in itertools.combinations(parts, 2):
+            nose.tools.assert_equal(set(), set(part1) & set(part2))
```
ysitu: What about exhaustiveness?

nxmetis/tests/test_metis.py
```diff
@@ -17,59 +20,70 @@ def make_cycle(n):
...
+        # The two separator nodes must not be present in the
+        # two bisected chains
+        nose.tools.ok_(sep[0] not in part1)
+        nose.tools.ok_(sep[0] not in part2)
+        nose.tools.ok_(sep[1] not in part1)
+        nose.tools.ok_(sep[1] not in part2)
+        
+        # The lists should be exhaustive with the node list of the Graph
+        nose.tools.ok_(set(sep) | set(part1) |
+                       set(part2) == set(self.G))
```
ysitu: Disjointedness?

nxmetis/tests/test_metis.py
OrkoHunter: Yes. Added them [L#78](https://github.com/networkx/networkx-metis/pull/44/files#diff-fbfe1ddde32d89fa89c2191c21aa9d6dR78) and [L#77](https://github.com/networkx/networkx-metis/pull/44/files#diff-fbfe1ddde32d89fa89c2191c21aa9d6dR77)

PR #47, opened 2015-07-20
Drop x64 testing on appveyor

### LINKED ISSUE #41: Fix AppVeyor CI

OrkoHunter:
Fixes #41

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,33 +14,20 @@ environment:
       PYTHON_VERSION: "2.7.9"
       PYTHON_ARCH: "32"
 
-    - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.9"
-      PYTHON_ARCH: "64"
-
     - PYTHON: "C:\\Python32"
       PYTHON_VERSION: "3.2.5"
       PYTHON_ARCH: "32"
-
-    - PYTHON: "C:\\Python32-x64"
-      PYTHON_VERSION: "3.2.5"
-      PYTHON_ARCH: "64"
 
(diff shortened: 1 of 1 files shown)


ysitu:
Please add a comment explaining why you need to do this and what it takes to reenable x64 testing.

OrkoHunter:
First, the default install of msvc does not support 64-bit tools, nevertheless the compiler for x64 is always present in the system with VS express. So, setuptools cannot work with a 64-bit Python install. However it *does* work for the default 32-bit arch.

This is an unresolved bug related to building a python package with msvc on x64 Windows arch. https://bugs.python.org/issue7511

However, there are temporary fixes for the build to work. [Here](http://poquitopicante.blogspot.in/2012/11/python-x64-package-extensions-with-pip.html) is one little *bad* method of doing it manually. However, there is a better way of the same. The reason `vcvarsall.bat` is complaining that it can't find x64 compilers even if they are correctly installed in the Visual Studio VC folder, is that the Paths set inside the file are wrong. One can change the Paths according to their drive and folder locations and the build should work just fine. Here is an [example patch](http://poquitopicante.blogspot.in/2014/04/fix-vcvarsallbat-to-install-python-27.html)

Now, Over a CI (AppVeyor) environment, I think it's not easy enough to apply the no-robust fix to make the x64 builds run. Rather we stop testing them over CI.

ysitu:
Sorry for being vague. I meant a comment in `appveyor.yml`. Now that you have typed that up, you can add a line in the file to refer to this PR.

OrkoHunter:
No problem. I added couple lines in the file.

PR #49, opened 2015-07-24
Some documentation changes

### LINKED ISSUE #48: Documentation issues
* https://networkx-metis.readthedocs.org/en/latest/reference/nxmetis.html lists the wrapper functions (`node_nested_dissection` and friends) but provides no links to their documentation.
* Documentation of each individual enum type contains two separate tables, one generated from the class docstring and one generated from the presence of enum values. We should suppress one of them (and also the documentation of `__init__`, which is not very helpful). If we keep the docstring table, we need to add a header like

```
Attributes
----------
<table code follows here...>
```

OrkoHunter:
Fix #48 

* Corrected the reference link
* Added MathJax extension

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -33,6 +33,7 @@ extensions = [
     'sphinx.ext.autosummary',
     'sphinx.ext.autodoc',
     'sphinx.ext.napoleon',
+    'sphinx.ext.mathjax',
     'sphinx.ext.intersphinx',
     'sphinx.ext.viewcode',
 ]
@@ -297,3 +298,7 @@ texinfo_documents = [
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'https://docs.python.org/': None}
+
+default_role = 'math'
+# Load MathJax into the webpage by including its main JavaScript file into the page
+mathjax_path = 'https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML'
--- a/doc/source/reference/nxmetis.rst
+++ b/doc/source/reference/nxmetis.rst
@@ -8,7 +8,7 @@ METIS Wrapper
 METIS
 -----
 
-.. automodule:: nxmetis.metis
+.. automodule:: nxmetis
 .. autosummary::
    :toctree: generated/
 
--- a/doc/source/reference/nxmetis.rst
+++ b/doc/source/reference/nxmetis.rst
@@ -2,8 +2,6 @@
 METIS Wrapper
 *************
 
-.. automodule:: nxmetis
-
 
 METIS
 -----
@@ -15,7 +13,6 @@ METIS
    node_nested_dissection
    partition
    vertex_separator
-   MetisOptions
 
 Enums
 -----
@@ -32,7 +29,6 @@ Enums
    MetisRType
    MetisNumbering
    MetisDbgLvl
-   MetisRStatus
 
 Types
 -----
@@ -42,12 +38,3 @@ Types
    :toctree: generated/
 
    MetisOptions
-
-Exceptions
-----------
-
-.. automodule:: nxmetis.exceptions
-.. autosummary::
-   :toctree: generated/
-
-   MetisError


OrkoHunter:
The changes can be seen here on [my fork on readthedocs](http://orkohunter-networkx-metis.readthedocs.org/en/doc-fixes/reference/generated/nxmetis.enums.MetisPType.html#nxmetis.enums.MetisPType)

nxmetis/enums.py
```diff
@@ -8,10 +8,12 @@
...
-        ==== =================================
+        Attributes
+        ----------
+        
+            rb :   Multilevel recursive bisectioning
+            kway : Multilevel `k`-way partitioning
+        
     """
 
     default, rb, kway = range(-1, 2)
```
ysitu: If this still does not work, what about

```python
class MetisPTye(enum.IntEnum):
    """Partitioning method."""

    default = -1
    """Default partitioning method.""""

    rb = 0
    """Multilevel recursive bisectioning."""

    kway = 1
    """Multilevel `k`-way partitioning."""
```

?

ysitu:
Please also look into enabling MathJax on readthedocs so that math expressions marked by single backticks are properly rendered.

nxmetis/enums.py
OrkoHunter: Yeah, this would suppress one of them. What about the docstring for `__init__()`?

ysitu: We should simply not show `__init__()` in the documentation.

OrkoHunter: The attributes failed to render in this way. [Link](http://orkohunter-networkx-metis.readthedocs.org/en/doc-fixes/reference/generated/nxmetis.enums.MetisPType.html#nxmetis.enums.MetisPType.kway)

OrkoHunter:
I think we will have to use `.. autoclass:` instead of `.. automodule:` to use the option of disabling documentation of private functions (eg. `__init__`).

From http://sphinx-doc.org/ext/autodoc.html#directive-autoclass :

> **Options and advanced usage**
>
If you want to automatically document members, there’s a members option:
>
.. automodule:: noodle
   :members:
will document all module members (recursively), and
>
.. autoclass:: Noodle
   :members:
will document all non-private member functions and properties (that is, those whose name doesn’t start with _).
>
For modules, __all__ will be respected when looking for members; the order of the members will also be the order in __all__.

doc/source/conf.py
```diff
@@ -33,6 +33,7 @@
     'sphinx.ext.autosummary',
     'sphinx.ext.autodoc',
     'sphinx.ext.napoleon',
+    'sphinx.ext.mathjax',
```
ysitu: You also need a line like https://github.com/networkx/networkx/blob/master/doc/source/conf.py#L228 to point to always load MathJax on HTTPS.

ysitu:
Please rebase. I changed the docstring of `partition()` a little bit (added an extra space). That did not fix the rendering, though. `tpwgts` and `ubvec` still look wrong.

OrkoHunter:
Okay thanks. I'll rebase after you are done.

ysitu:
I am already done. The change was small enough that I did it directly on GitHub.

doc/source/conf.py
OrkoHunter: Added the path to the main JavaScript file of MathJax.

ysitu:
According to the [napoleon doc](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/index.html#docstring-sections), it should support "Attributes" sections. We can disable the auto generated doc for class members and produce the documentation directly from docstrings.

ysitu:
The same doc also says that napoleon requires running `sphinx-apidoc`, which seems not to be a replacement for `sphinx-build`. The latter is still needed as I read it.

OrkoHunter:
From http://sphinx-doc.org/man/sphinx-apidoc.html :

> `sphinx-apidoc` generates source files that use `sphinx.ext.autodoc` to document all found modules. If any modules have side effects on import, these will be executed by `autodoc` when `sphinx-build` is run.

I guess we should give a try with `sphinx-apidoc`. For that purpose, we have to change the `Makefile`. Am I right?

OrkoHunter:
After using `Attributes`, I think that the table was bit prettier than how `Attributes` lists the parameters.

ysitu:
`sphinx-apidoc` is not necessary. It is used to generate `nxmetis.rst` and friends. But you already have an (arguably better) handwritten copy.

doc/source/conf.py
```diff
@@ -297,3 +298,6 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'https://docs.python.org/': None}
+
+# Load MathJax into the webpage by including its main JavaScript file into the page
+mathjax_path = 'https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML'
```
ysitu: Also add `default_role = 'math'` so that contents between single backticks are parsed as math.

doc/source/conf.py
```diff
@@ -297,3 +298,11 @@
...
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'https://docs.python.org/': None}
+
+default_role = 'math'
+# Load MathJax into the webpage by including its main JavaScript file into the page
+mathjax_path = 'https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML'
+
+# napoleon settings
+napoleon_include_private_with_doc = False
+napoleon_include_special_with_doc = True
```
OrkoHunter: According to the [doc](http://sphinx-doc.org/latest/ext/napoleon.html#confval-napoleon_include_special_with_doc), after setting `napoleon_include_special_with_doc` to true, special members like `__init__` without docstring are not included in the documentation.

doc/source/reference/nxmetis.rst
```diff
@@ -8,7 +8,7 @@ METIS Wrapper
 METIS
```
ysitu: Remove Line 5.

doc/source/conf.py
ysitu: But it seems not to be the case in your readthedocs build.

nxmetis/enums.py
ysitu: Hmm... This works for me.

ysitu:
Putting docstrings beneath enum values works for me, but I do not know how to get rid of the `__init__()` functions.

doc/source/reference/nxmetis.rst
ysitu: Also remove Line 18.

doc/source/conf.py
OrkoHunter: Yeah.. It should but it won't. Wow! :)

nxmetis/enums.py
OrkoHunter: Sorry I didn't understand. Does this work on your fork of readthedocs? Or did you build the doc locally?

ysitu: I was building locally.

OrkoHunter:
The problem which we are facing is already solved and mentioned in using [.. autoclass::](http://sphinx-doc.org/ext/autodoc.html#directive-autoclass). And about the duplication of table, so let's keep it as you suggested (keeping docstrings below enum values.)

nxmetis/enums.py
```diff
@@ -6,122 +6,154 @@
...
 
-        ====== ======================================================
-        grow   Grow a bisection using a greedy strategy
-        random Compute a bisection at random followed by a refinement
-        edge   Derive a separator from an edge cut
-        node   Grow a bisection using a greedy node-based strategy
-        ====== ======================================================
-    """
+    Attributes
+    ----------
```
ysitu: Remove these two lines.

nxmetis/enums.py
```diff
@@ -6,122 +6,154 @@
...
-        cut Edge-cut minimization
-        vol Total communication volume minimization
-        === =======================================
-    """
+    """Type of objective."""
 
-    default, cut, vol = range(-1, 2)
+    default = -1
+    """Default type of objective"""
+    cut = 0
```
ysitu: Add blank lines between attributes, otherwise it is hard to tell which attribute each docstring belongs to.

nxmetis/enums.py
```diff
@@ -6,122 +6,154 @@
...
+
+    conninfo = 128
+    """Display information related to the minimization of
+    subdomain connectivity"""
+
+    contiginfo = 256
+    """Display information related to the elimination of
+    connected components"""
+
+    memory = 2048
```
ysitu: Let's remove this. It is undocumented and ignored by METIS.

ysitu:
Please end all docstrings with periods.

doc/source/reference/nxmetis.rst
```diff
@@ -2,20 +2,17 @@
...
 .. autosummary::
    :toctree: generated/
 
    node_nested_dissection
    partition
    vertex_separator
-   MetisOptions
 
 Enums
 -----
```
ysitu: Shall we remove `MetisRStatus` from the documentation? It is used only internally for relaying return codes and error message.

OrkoHunter: Yes, there is no need of `MetisRStatus` in the documentation.

ysitu:
Did you publish your changes to your readthedocs fork? I am still seeing the old rendering.

OrkoHunter:
readthedocs automatically builds on every commit. Builds for this branch are published here.
http://orkohunter-networkx-metis.readthedocs.org/en/doc-fixes/

OrkoHunter:
I just wiped the version on readthedocs and triggered a fresh build.

OrkoHunter:
No changes though.

ysitu:
Well, this is really beyond me. Perhaps we need to ditch autodoc? @hagberg @chebee7i, do you have any suggestions?

OrkoHunter:
The final solution can be rewriting `nxmetis.rst` with the docstrings from the module.

OrkoHunter:
So, I believe we drop the use of `autodoc` if this is blocking the release.

ysitu:
That makes sense as documentation is the last blocker.

OrkoHunter:
I pushed the changes.

In the [toctree](http://orkohunter-networkx-metis.readthedocs.org/en/custom/reference/nxmetis.html#enums), I don't know how does `autodoc` generate a table. Efforts like

```
   +----------------+------------------------------------------+
   | MetisPType     | Partitioning method                      |
   +----------------+------------------------------------------+
```
and
```
==========  ====================
MetisPType  Partitioning method.
==========  ====================
```
render nothing.

This
```
Partitioning method. <MetisPType>
```
just replaces the enum with description which is far away from what we want.

OrkoHunter:
Please see http://orkohunter-networkx-metis.readthedocs.org/en/doc-fixes/

ysitu:
Let's proceed with this step by step. Can you move the changes to `enums.py` to a separate PR?

doc/source/conf.py
ysitu: These two lines do not help because `napoleon` registers a `autodoc-skip-member` handler that explicitly exempts `__init__`.

ysitu:
After much trial-and-error, this seems to reliably get rid of `__init__()` for me:

1. Put

```rst
{{ fullname }}
{{ underline }}

.. currentmodule:: {{ module }}

.. autoclass:: {{ objname }}

   {% block attributes %}
   {% if attributes %}
   .. rubric:: Attributes

   .. autosummary::
   {% for item in attributes %}
      ~{{ name }}.{{ item }}
   {%- endfor %}
   {% endif %}
   {% endblock %}
```

in `source/_templates/enum.rst`.

2. Use the template by adding `:template: enum.rst` to `autosummary` in the "Enums" section:

```diff
 .. automodule:: nxmetis.enums
 .. autosummary::
    :toctree: generated/
+   :template: enum.rst
```

3. Remove `source/reference/generated` and rebuild.

OrkoHunter:
#50 

Yes this really removes `__init__.py`. See the results [here](http://orkohunter-networkx-metis.readthedocs.org/en/enum-doc/reference/generated/nxmetis.enums.MetisPType.html#nxmetis.enums.MetisPType)

ysitu:
Now can you move the changes to `nxmetis.rst` and `conf.py` to a separate PR so that we can merge them?

OrkoHunter:
This PR is especially for the changes to `nxmetis.rst` and `conf.py`. I'll remove the extra `.rst` files. And I think the duplication of table seems to be resolved by [this commit](https://github.com/OrkoHunter/networkx-metis/commit/2b75b0e9c48bee24d04665091d0a40be576392e0). See [this branch](http://orkohunter-networkx-metis.readthedocs.org/en/test/reference/generated/nxmetis.enums.MetisPType.html#nxmetis.enums.MetisPType).

ysitu:
Please revert the extra `.rst` files so that we can merge.

doc/source/reference/nxmetis.rst
```diff
@@ -50,4 +46,4 @@ Exceptions
 .. autosummary::
    :toctree: generated/
 
-   MetisError
+   MetisError
```
ysitu: Remove `MetisError`. We are not exposing this type.

OrkoHunter:
It's okay now.

doc/source/reference/nxmetis.rst
```diff
@@ -49,5 +45,3 @@ Exceptions
 .. automodule:: nxmetis.exceptions
 .. autosummary::
    :toctree: generated/
-
-   MetisError
```
ysitu: Remove this section entirely now that it is empty.

OrkoHunter: Yeah that makes more sense.

PR #51, opened 2015-08-03
Remove the duplication of Attributes table in documentation of enums

OrkoHunter:
http://orkohunter-networkx-metis.readthedocs.org/en/test

### DIFF
--- a/doc/source/_templates/enum.rst
+++ b/doc/source/_templates/enum.rst
@@ -9,9 +9,5 @@
    {% if attributes %}
    .. rubric:: Attributes
 
-   .. autosummary::
-   {% for item in attributes %}
-      ~{{ name }}.{{ item }}
-   {%- endfor %}
    {% endif %}
    {% endblock %}
--- a/nxmetis/enums.py
+++ b/nxmetis/enums.py
@@ -6,152 +6,122 @@ __all__ = ['MetisPType', 'MetisObjType', 'MetisCType', 'MetisIPType',
 
 @enum.unique
 class MetisPType(enum.IntEnum):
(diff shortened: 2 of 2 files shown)


OrkoHunter:
Are we not interested in this?
I see nothing rendered [here](http://networkx-metis.readthedocs.org/en/latest/reference/generated/nxmetis.enums.MetisPType.html#nxmetis.enums.MetisPType) in the one Attributes table.

ysitu:
I want to see if fixing #52 also fixes this.

OrkoHunter:
The Attributes table is still without any content.

ysitu:
Hmm... I triggered a doc rebuild, and now docstrings of enum values show up in the tables.

OrkoHunter:
That looks great!

ysitu:
After that I rebuilt several times. Sometimes it worked, sometimes it didn't. Still not sure why that change from build to build.

OrkoHunter:
Okay, so can we make new branch from here on and refer 'latest' on readthedocs to that branch? This will  prevent from re-building several times.

ysitu:
Not sure what you mean. I manually triggered the rebuild on the readthedocs site.

OrkoHunter:
Because the documentation sometimes work and sometimes didn't, I suggest to create a new branch in the repository like `stable` which would be in the sense of "stable-documentation". And on [readthedocs](https://readthedocs.org/dashboard/networkx-metis/versions/), we make the 'latest' version inactive and the 'stable' version active.

This is just a simple hack to stop continuous builds over the active version on readthedocs which might disturb the documentation.

PR #53, opened 2015-08-03
Put function definitions separate from nxmetis/__init__.py

### LINKED ISSUE #52: Make nxmetis importable even if nxmetis._metis is not built
When building documentation, `autodoc` attempts to import `nxmetis`. If `nxmetis._metis` is not built, import will fail and cause errors. One solution is to modify imports of `_metis` in `__init__.py` and `types.py` to read

```python
try:
    from nxmetis import _metis
except ImportError:
    pass
```

But I think that it is probably better to create a `metis.py` that imports stuff from `_metis`

```python
try:
    from nxmetis._metis import node_nd, ...
except ImportError:
    def node_nd(*args, **kargs):
        raise ...
    ...
```

and import `metis` instead of `_metis` in `__init__.py` and `types.py`.

OrkoHunter:
Fix #52

### DIFF
--- a/nxmetis/metis.py
+++ b/nxmetis/metis.py
@@ -1,0 +1,24 @@
+"""
+    This module is a wrapper of the ``_metis`` extension generated
+    from ``_metis.pyx``. This is to let sphinx import the module for
+    documentation builds with or without the ``_metis`` extension.
+"""
+try:
+    from nxmetis import _metis
+except ImportError:
+    def node_nd(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+
+    def part_graph(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+
+    def compute_vertex_separator(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+
+    def set_default_options(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+else:
+    node_nd = _metis.node_nd
+    part_graph = _metis.part_graph
+    compute_vertex_separator = _metis.compute_vertex_separator
+    set_default_options = _metis.set_default_options
--- a/nxmetis/__init__.py
+++ b/nxmetis/__init__.py
@@ -17,7 +17,7 @@ import six
 
 from nxmetis import enums
 from nxmetis import exceptions
-from nxmetis import _metis
+from nxmetis import metis
 from nxmetis import types
 
 __all__ = ['node_nested_dissection', 'partition', 'vertex_separator',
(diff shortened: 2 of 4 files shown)


nxmetis/metis.py
```diff
@@ -0,0 +1,317 @@
...
+import networkx as nx
+import six
+
+from nxmetis import enums
+from nxmetis import exceptions
+from nxmetis import types
+try:
+    from nxmetis import _metis
+except ImportError:
+    pass
```
OrkoHunter: I couldn't make it like below as you suggested because I believe we do not import `node_nd` etc. from `_metis` rather use it like `_metis.node_nd`
```py
try:
    from nxmetis._metis import node_nd, ...
except ImportError:
    def node_nd(*args, **kargs):
        raise ...
    ...
```

ysitu: You can do this:

```python
try:
    from nxmetis import _metis
except ImportError:
    def node_nd(...):
        ...
    ...
else:
    node_nd = _metis.node_nd
    ...
```

nxmetis/metis.py
```diff
@@ -0,0 +1,326 @@
...
+    def node_nd(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+    def part_graph(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+    def compute_vertex_separator(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+else:
+    node_nd = _metis.node_nd
+    part_graph = _metis.part_graph
+    compute_vertex_separator = _metis.compute_vertex_separator
```
OrkoHunter: This will raise an exception if a function is called from `_metis` and if `_metis` was not built.

ysitu:
You do not have to rename `__init__.py`. In particular, you do not need to worry about circular imports. Resolving the headache with that was exactly reason why we went with importing only modules and nothing else. All you need here is to provide a thin wrapper for `_metis` named `metis` and use it `__init__.py` and `types.py`.

OrkoHunter:
I hope this is okay now.

nxmetis/metis.py
```diff
@@ -0,0 +1,16 @@
+try:
```
ysitu: Add a comment on why we need to do this.

nxmetis/metis.py
```diff
@@ -0,0 +1,16 @@
...
+    from nxmetis import _metis
+except ImportError:
+    def node_nd(*args, **kwargs):
+        raise Exception("NetworkX-METIS not installed!")
+    def part_graph(*args, **kwargs):
```
ysitu: Add a blank line between functions.

OrkoHunter: Done.

nxmetis/metis.py
```diff
@@ -0,0 +1,24 @@
+"""
+    This module is a wrapper of the ``_metis`` extension generated
+    from ``_metis.pyx``. This is required in the situations when
+    functions are called before the build of ``_metis``.
+"""
```
ysitu: No. This is specifically to allow Sphinx to import the module during doc builds. No function calls are allowed.

OrkoHunter: Yes I messed up between the import and function call. Changed now.

PR #55, opened 2015-08-09
Minor doc fixes

SanketDG:
This fixes the warnings that appear in https://readthedocs.org/builds/networkx-metis/3214597/ and some minor ambiguities.

Also, I really doubt the fact that the docs have to follow the exact format of it's parent project. Do we really need to have a section titled "The Python Programming Language"? This is a fairly new project, so removing "Who uses Networkx-METIS" seems natural to me.

All in all, I feel a lot of parts need to be changed in the docs where we are following NetworkX, but largely going out of path what this project actually is. It's a good practice to follow the parent project, but NetworkX and NetworkX-METIS are very different projects from a high-level point of view IMHO.

### DIFF
--- a/doc/source/bibliography.rst
+++ b/doc/source/bibliography.rst
@@ -1,0 +1,11 @@
+..  -*- coding: utf-8 -*-
+
+Bibliography
+============
+
+.. [Langtangen04] H.P. Langtangen, "Python Scripting for Computational
+    Science.", Springer Verlag Series in Computational Science and
+    Engineering, 2004.
+
+.. [Martelli03]  A. Martelli, "Python in a Nutshell", O'Reilly Media
+   Inc, 2003.
--- a/doc/source/download.rst
+++ b/doc/source/download.rst
@@ -5,7 +5,7 @@ Download
 Software
 ~~~~~~~~
 
-Source and binary releases: https://pypi.python.org/pypi/networkx/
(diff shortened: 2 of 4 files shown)


OrkoHunter:
Frankly speaking, the `overview` part is simply following the NetworkX documentation and I see no harm in that. I think NetworkX holds all the grounds for its add-ons which simply provide APIs for one or more functionality. That's why I don't like calling both of them two "different projects". However, there are many redundant information in the overview, which sure can be removed e.g. "Who uses NetworkX-METIS".

SanketDG:
I will want to amend my last sentence now, I believe they are different because of their scope and size, NetworkX is huge in size and provides large amount of functions and classes in its API. NXMETIS provides just a subset of the functionality that NetworkX provides.

From `overview`: 

I find the whole section of the _Python Programming Language_ to be redundant. NetworkX came up in 2005, when adoption of Python was in its infancy. For a 2015 add-on of a fairly popular python package,the information is redundant. Also we are suggesting a book from 2003. Most people will only discover NetworkX-METIS after NetworkX, also stating the pros of python in 2015 is unnecessary.

> Information on NetworkX development is found at the NetworkX Developer Zone at Github

I wouldn't use the word "Developer Zone" but this can be taken with a grain of salt.

OrkoHunter:
This is ready to go.

PR #61, opened 2019-10-31
Now runs with new networkx

Lamaun:
The change of networkx graph data structure broke this package.

### DIFF
--- a/nxmetis/tests/test_metis.py
+++ b/nxmetis/tests/test_metis.py
@@ -24,7 +24,7 @@ class TestMetis(object):
         self.node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
                           1, 2, 3, 4, 5, 6]
         self.G = nx.Graph()
-        self.G.update(nx.path_graph(self.node_list))
+        nx.add_path(self.G,self.node_list)
         self.G.add_edge(self.node_list[-1], self.node_list[0])
 
     def test_node_nested_dissection_unweighted(self):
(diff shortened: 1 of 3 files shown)


nxmetis/tests/test_metis.py
```diff
@@ -24,7 +24,7 @@ def setUp(self):
         self.node_list = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
                           1, 2, 3, 4, 5, 6]
         self.G = nx.Graph()
-        self.G.add_path(self.node_list)
+        self.G.update(nx.path_graph(self.node_list))
```
dschult: ```self.G.add_path(stuff)``` -- > ```nx.add_path(self.G, stuff)``` is a good way to add a path.

dschult:
This repository hasn't been actively updated in a while.
Probably the python versions tested on Travis should match the current networkx versions.
It is possible that other deprecated functionality from NetworkX could cause problems too. 
Thanks for doing this!

Lamaun:
I haven't worked with travis yet, should we replace travis.yml with something similar to https://github.com/networkx/networkx/blob/master/.travis.yml?

dschult:
Can you try a more simple change first?  I'm hoping it doesn't require the more heavy lifting.

Change it to python versions 3.6, 3.7, 3.8, pypy3 and let's see what Travis says.
Thanks!

Lamaun:
Seems like that works

dschult:
Excellent -- anything else before we merge?

Lamaun:
Not from my side. Could we get this into pypi?

dschult:
I'll look into getting it into pypi.

PR #64, opened 2019-11-05
cleanup nx decorators

gwendolynbrook:
Only use one decorator to validate structure of nx graph.

https://github.com/networkx/networkx/blob/master/networkx/utils/decorators.py#L52


dschult:
Be careful.  Two sequential decorators connect as an "OR" while One decorator with two entries connect as an "AND". So, you have changed it from "not allowing directed graphs while also not allowing multigraphs" to "not allowing directed multigraphs".

